PR #2, opened 2015-03-12
Better packaging.

cadesalaberry:


bsansouci:
Hey thanks for the PR. I don't really like putting a source file into a folder called bin which usually is used for compiled binaries. If anything we could move time.js to a lib/ folder.

Schmavery:
Not sure what the point of adding files we don't have to the gitignore just for the sake of using some default, either.

cadesalaberry:
Sorry, I meant to put it in the `lib` folder.

As for the `.gitignore`, it is good practice to do it at the begining of the project to avoid accidental commits.

PR #9, opened 2015-04-30
Added option of logging in via json

Tankenstein:
This creates the option of using login() with a javascript object, instead of reading from file. Convenient, since you're probably parsing a file anyways, no need to do it twice.


Tankenstein:
Great! Next up - conditional logging :)

PR #24, opened 2015-06-30
Fix non-page login

zacharyliu:
When logging in with a standard account, the `api` variable was not properly passed to the callback.


bsansouci:
:+1:

Schmavery:
U DA MAAAAN

PR #25, opened 2015-06-30
Fixed README error

mjkaufer:
You were passing a string to the login object, when in actuality, you had to pass an actual object. Fixed the error by using `fs` to read from the config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,21 @@ npm install facebook-chat-api
 ## Example Usage
 ```javascript
 var login = require("facebook-chat-api");
+var fs = require("fs");
 
 // Create simple echo bot
-login("config.json", function(err, api) {
-    if(err) return console.error(err);
-  
-    api.listen(function callback(err, message) {
-        api.sendMessage(message.body, message.thread_id);
-    });
-});
+fs.readFile('config.json', function(err, data){
+	var json = JSON.parse(data)
+
+	login(json, function(err, api) {
+	    if(err) return console.error(err);
+	    
+	    api.listen(function callback(err, message) {
+	        api.sendMessage(message.body, message.thread_id);
+	    });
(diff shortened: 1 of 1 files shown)


bsansouci:
Good catch! Do you mind fixing all of the logins? I think we use it in most examples. Maybe using env variables because it would be shorter...

mjkaufer:
Sure. I'll probably do a `fs.readFileSync` in the interest of not having the code grow sideways faster than vertically. Alternately, I can implement the file reading in `index.js` so the current documentation is valid. Either way, I'm going to close this, and we can continue discussing here: https://github.com/Schmavery/facebook-chat-api/pull/27

bsansouci:
If you want to fix the Readme, let's keep the discussion here instead of polluting other PRs.
No don't add it in index.js, we removed it because there are too many ways to load that data. It doesn't make sense for us to implement only a few of them nor does it make sense to implement all possible ways, so we just don't implement them.

readFileSync works for me :)

mjkaufer:
Sounds good. I updated the rest of the examples.

Schmavery:
Good catch, we should have read over the README closer after removing that functionality.
I think it's important to keep the examples short and to the point, while still being functional working (copy pasteable) examples.

The filename was used in order to shorten the examples, but relying on a file introduces a black box into the example where it's not immediately clear what data should be in the file.

I think it would be clearer and more concise to provide examples in this form:
```js
login({email: "FB_EMAIL", password: "FB_PASSWORD"}, function callback (err, api) {
    // stuff
});
```

mjkaufer:
Sounds good. I just added that.

bsansouci:
Thanks a lot :+1:

mjkaufer:
No problem

Issue #26, opened 2015-06-30
getUserId Error


mjkaufer:
Hi,

Calling `api.getUserId` breaks the program. I tried calling the method itself manually and got an error, and calling `api.sendDirectMessage`, which uses `api.getUserId`, throws the same error.

The error is as follows

```
WARN Warning - sendDirectSticker and sendDirectMessage are deprecated. 
ERR! Error in getUserId TypeError: Cannot read property '0' of undefined
ERR! Error in getUserId     at /Users/matt/Projects/node_modules/facebook-chat-api/src/getUserId.js:21:25
ERR! Error in getUserId     at tryCatcher (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:24:31)
ERR! Error in getUserId     at Promise._settlePromiseFromHandler (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:454:31)
ERR! Error in getUserId     at Promise._settlePromiseAt (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:530:18)
ERR! Error in getUserId     at Promise._settlePromises (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:646:14)
ERR! Error in getUserId     at Async._drainQueue (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:182:16)
ERR! Error in getUserId     at Async._drainQueues (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:192:10)
ERR! Error in getUserId     at Immediate.Async.drainQueues [as _onImmediate] (/Users/matt/Projects/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:15:14)
ERR! Error in getUserId     at processImmediate [as _immediateCallback] (timers.js:367:17)
ERR! Error in getUserId  [TypeError: Cannot read property '0' of undefined]
[TypeError: Cannot read property '0' of undefined]

```

mjkaufer:
See PR https://github.com/Schmavery/facebook-chat-api/pull/27

PR #27, opened 2015-06-30
Fixed resData.entries error

mjkaufer:
Fixed the following issue, https://github.com/Schmavery/facebook-chat-api/issues/26, by changing references to `resData.entries` to `resData.payload.entries`

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ Given a person's full name will do a Facebook Graph search and return all the id
 __Arguments__
 
 * `name` - A string being the name of the person you're looking for.
-* `callback(err, obj)` - A callback called when the search is done (either with an error or with the resulting object). `obj` contains and array called entries which contains all the users that facebook graph search found, ordered by "importance".
+* `callback(err, obj)` - A callback called when the search is done (either with an error or with the resulting object). `obj` is an array which contains all of the users that facebook graph search found, ordered by "importance".
 
 __Example__
 
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ login('config.json', function(err, api) {
         if(err) return callback(err);
         
         // Send the message to the best match (best by Facebook's criteria)
-        var thread_id = data.payload.entries[0].uid;
+        var thread_id = data[0].uid;
         api.sendMessage(msg, thread_id);
     });
 });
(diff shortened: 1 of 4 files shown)


bsansouci:
Thanks for this fix! I feel like it doesn't make much sense for getUserId to return that big object. Maybe it should return `resData.payload.entries` instead, because that's the only thing that really matters. What do you think?

mjkaufer:
Yeah, I think `resData` has a bunch of junk. Really, the only potentially useful info is in payload.

bsansouci:
Ok cool, then if you can make getUserInfo only return `resData.payload.entries` we can merge this PR

mjkaufer:
That said, lots of methods return `resData`, which I don't think have entries objects, so it would make a bit of an inconsistency across methods, so maybe returning `resData` is the best option

mjkaufer:
Let me know, and I'll update the method and fix the README for this stuff and the login feature. 

Also, do we want to read the `config.json` file from the `login` function itself, or should we just call a read function and pass the object to `login`?

bsansouci:
> That said, lots of methods return resData, which I don't think have entries objects, so it would make a bit of an inconsistency across methods, so maybe returning resData is the best option

I don't know which methods you're referring to, but that sounds like it shouldn't be the case. Can you point out to me which ones you're thinking of? I think it makes more sense to avoid exposing FB directly because they don't mean a lot without context. @Schmavery thoughts?

As I said in the other PR, we should talk about the readme stuff over there.

mjkaufer:
I implemented the changes you wanted!

bsansouci:
Cool thanks :)
Will merge once you update sendDirectMessage and sendDirectSticker

mjkaufer:
I'm not sure what I need to do to `sendDirectMessage` and `sendDirectSticker`... they both use payload now, don't they?

bsansouci:
Well `getUserId` already returns `resData.payload.entries`, so instead of `data.payload.entries[0].uid` you're gonna want `data[0].uid`

mjkaufer:
Oh, whoops

mjkaufer:
Everything's up to date. Thanks for the catch!

Schmavery:
One last thing - doesn't the README example for `getUserId()` still refer to `data.payload.entries[0].uid`?

mjkaufer:
Not anymore! :smile:

PR #35, opened 2015-07-08
Added way to get file attachment

webfreak7:
self explanatory. Pretty much takes care of any file attachments you throw at it. Tried adding photo attachments, but the url returned was super weird and didn't work (something along the lines of /ajax/mercury/attachements...). Will add if I find a way to do it.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -113,6 +113,20 @@ function formatMessage(m) {
         obj.file_url = originalMessage.attachments[i].url;
         break;
       }
+      if (originalMessage.attachments[i].attach_type === "photo"){
+        obj.type = "photo";
+        delete obj.body;
+        obj.name = originalMessage.attachments[i].name;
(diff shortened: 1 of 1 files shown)


blazeu:
there's ```preview_url```, ```thumbnail_url``` and ```hires_url``` for photo attachments and seems to be working fine if opened.

bsansouci:
When you say "Tried adding photo attachments" you mean upload photos? Because we're working on that, we'll publish soon.

I just tried to send myself a picture and the `type` was `"photo"` and not `"file"`. @blazeu is right, hires_url seems to be the full version. My guess is that if you do a post request (after being logged in) to http://facebook.com/{attachement.url}, you'll get the image.
 
Can make this work for picture too? I'll merge after that :)

webfreak7:
I added in support for photos and gifs.

Schmavery:
Awesome, thanks.

PR #38, opened 2015-07-10
Adding getCurrentUserId to API

MarkehMe:
This is a simple way to get the current REAL user ID.

The regular Facebook API gives a new ID per-app which was not helpful in my case, thankfully the cookies hold the delicious solution. 

    api.getCurrentUserId(function(userID) {
        console.log("Your REAL user ID is " + userID);
    });

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,7 +128,8 @@ function _login(email, password, loginOptions, callback) {
         'getFriendsList',
         'getUserInfo',
         'removeUserFromGroup',
-        'addUserToGroup'];
+        'addUserToGroup',
+        'getCurrentUserId'];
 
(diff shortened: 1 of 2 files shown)


Schmavery:
I suppose there's no reason why we can't provide this functionality, but it's actually a one-liner because we already have the data in the context that we pass to each function.

```
module.exports = function(mergeWithDefaults, api, ctx) {
  return function getCurrentUserId() {
    return ctx.userId;
  };
};
```

I'd imagine something like that would work.

MarkehMe:
Ill make a new pull request, that seems a lot less intensive then what I had going.

PR #39, opened 2015-07-10
Add api.sendTypingIndicator

blazeu:
Send typing indicator for other users in the chat to see.

The typing indicator will disappear automatically after 30 seconds or after sending a message (```api.sendMessage```).

However, this can be overridden by sending a request again, with the field ```typ``` being ```0``` instead of ```1```.
Still thinking the best idea to implement this tho.

Any suggestion?

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,7 +128,8 @@ function _login(email, password, loginOptions, callback) {
         'getFriendsList',
         'getUserInfo',
         'removeUserFromGroup',
-        'addUserToGroup'];
+        'addUserToGroup',
+        'typing'];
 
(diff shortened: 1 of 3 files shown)


Schmavery:
Am I correct in understanding that this is in order to send a indication to other users that the api is typing, rather than detecting other users who are typing?

blazeu:
Yes, this is for **sending** typing indicator.

Schmavery:
Maybe we can call it `sendTypingIndicator` and have it call the callback with a function that allows you to end the indicator early (with the `typ` field `0`).

bsansouci:
This looks good, thanks a lot :D

PR #40, opened 2015-07-12
Adding getCurrentUserId endpoint

MarkehMe:
Much simpler and less cookie orientated method to get the current user! 

Usage example:

    var UserID = api.getCurrentUserId();

*Note to self: explore more of the code before contributing!*


bsansouci:
Cool, thanks!

PR #41, opened 2015-07-12
Add getThreadHistory, getThreadHistoryAll and getThreadListAll

g0ral:


Schmavery:
Can we get any explanation of what you mean for this to do?  Does it simply fetch the chat history of a thread?

g0ral:
Yes. This method fetch latest messages of a thread. You have to give thread id, start and limit for this one. I didn't test it for max limit yet. 
I have wrote this to download all facebook threads and theirs messages to database. I will write later getThreadAll function to download all messages from a thread.

Schmavery:
Can you match the `if (end <= start) end = start + 20;` behaviour from `getThreadList` and call the function `getThreadHistory`?  

I also feel like it would be nice if `getThreadAll` was just an overloaded version of this method, with no `start` and `end` specified.  

Can you rename the `threadFbid` argument to `thread_id` for some more consistency (though our and Facebook's lack of naming consistency is duly noted)?

It would be great to be able to get an entry in the README for this function.

I'm trying to decide if the messages should be processed before we return them.  @bsansouci do you think it will work to map `utils.formatMessage` over the result before returning?

bsansouci:
Yes, for the sake of consistency. We need to make sure those messages are of the same shape.

Schmavery:
Not sure if the message history includes events, but if it does, those will also need to be processed with `formatEvent` or stripped out according to `setOptions`.

g0ral:
I have changed things you requested but I can't add `utils.formatMessage` because those messages have different shape. There is no `sender_*` variables. Example:

```javascript
{ message_id: 'mid.1436508866652:6d0bafd2e680648024',
  threading_id: null,
  offline_threading_id: '3469248605',
  author: 'fbid:100004402814308',
  author_email: '100004402814308@facebook.com',
  timestamp: 1436508866713,
  timestamp_absolute: 'piątek',
  timestamp_relative: 'piąt.',
  timestamp_datetime: 'piątek 08:14',
  timestamp_time_passed: 1,
  is_unread: false,
  is_forward: false,
  is_filtered_content: false,
  is_filtered_content_bh: false,
  forward_count: 0,
  forward_message_ids: [],
  source: 'source:titan:web',
  source_tags: [],
  is_spoof_warning: false,
  folder: 'inbox',
  thread_fbid: '100000213419114',
  other_user_fbid: 100000213419114,
  body: 'wiadomość 115',
  html_body: null,
  subject: null,
  has_attachment: false,
  attachments: [],
  raw_attachments: null,
  ranges: [],
  thread_id: 'mid.1435326172529:58a6982888d5a96185',
  action_id: null,
  action_type: 'ma-type:user-generated-message' }
```

g0ral:
Are you sure `start`, `end` and `if (end <= start) end = start + 20;` is good for offset and limit? With `start` and `end` when I write `getThread(id, 10, 15)` I should get messages 10-15 but instead I will get messages 10-25.

bsansouci:
Good point, totally got that wrong. Can you change it in the PR and we'll fix the other ones.
It should be `offset = start` and `limit = end - start` right?

Schmavery:
Can we update `utils.formatMessage` to be flexible enough to handle this?

g0ral:
There is no `sender_name` only fb id so it will be hard to use `utils.formatMessage` and fetch all data it should provide.

bsansouci:
Mmh ok. I'm still in favor of being consistent. What do you think of having those methods (etThreadHistory, getThreadHistoryAll and getThreadListAll) get all the info about those messages to be able to put through `utils.formatMessage`? I can't think of a use case off the top of my head where you'd want only the message id or body and not it's sender info...

bsansouci:
Mmh ok. I'm still in favor of being consistent. @g0ral, @Schmavery What do you think of having those methods (etThreadHistory, getThreadHistoryAll and getThreadListAll) get all the info about those messages to be able to put through utils.formatMessage? I can't think of a use case off the top of my head where you'd want only the message id or body and not it's sender info...

Schmavery:
I definitely agree that they should pull all relevant data.  If it's just the sender_name we need, it should be possible to get that fairly intelligently using existing api functionality.

blazeu:
This doesn't work on group chat.

bsansouci:
@g0ral Hey man, do you mind closing this PR and opening it again against the pre-release branch? We'll fix the `send_name` issue.

Schmavery:
I'm going to close this PR as I just opened it against the pre-release branch myself.

Issue #43, opened 2015-07-13
Support upload


Schmavery:
We don't yet support file upload.  It would be nice to condense this into as few library functions as possible.  My hunch is that it will be possible to have one `api.sendAttachment` function that will support both regular and animated images as well as standard file upload.

If anyone wants to take a stab at this, feel free :smile:

bsansouci:
Fixed in #44

PR #44, opened 2015-07-13
Add attachment upload

blazeu:
This should solve #43 

It's a little different than @Schmavery has had envisioned. Instead of using different function to allow attachment upload, i'll be using ```api.sendMessage``` by passing an object containing the **body** (string) and **attachment** (readable stream or an array of readable stream). I think this is much simpler.

```
var msg = {
  body: 'message body',
  attachment: fs.createReadStream(__dirname + '/image.jpg')
}

api.sendMessage(msg, message.thread_id)
```

```api.sendMessage``` will still support passing string in the first argument, so the current example in the documentation still works.

As for the ```api.uploadAttachment```,
it's a function to *upload* the attachment, it will return the attachment id to be passed in the message form. The ```api.sendMessage``` will invoke this first before sending the message, if the new message contain an attachment.

i'd recommend making ```api.sendSticker``` like this too. Since the form data is similar, you should avoid repeating yourself.

### DIFF
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -59,7 +59,7 @@ module.exports = function(mergeWithDefaults, api, ctx) {
       form['message_batch[0][gif_ids]'] = [];
       form['message_batch[0][file_ids]'] = [];
 
-      if (!utils.getType(msg.attachment) !== 'Array') {
+      if (utils.getType(msg.attachment) !== 'Array') {
         msg.attachment = [msg.attachment]
       }
 
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -7,8 +7,8 @@ var log = require("npmlog");
 module.exports = function(mergeWithDefaults, api, ctx) {
   return function sendMessage(msg, thread_id, callback) {
     if(!callback) callback = function() {};
-    if(typeof msg !== "string" && typeof msg !== "object")
-      return callback({error: "Message should be of type string or object and not " + typeof msg + "."});
+    if(typeof msg !== "string" && !utils.checkType(msg, "Object"))
+      return callback({error: "Message should be of type string or object and not " + utils.getType(msg) + "."});
     if(typeof thread_id !== "number" && typeof thread_id !== "string")
       return callback({error: "Thread_id should be of type number or string and not " + typeof thread_id + "."});
 
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -7,13 +7,21 @@ var log = require("npmlog");
 module.exports = function(mergeWithDefaults, api, ctx) {
   return function sendMessage(msg, thread_id, callback) {
     if(!callback) callback = function() {};
-    if(typeof msg !== "string" && !utils.checkType(msg, "Object"))
-      return callback({error: "Message should be of type string or object and not " + utils.getType(msg) + "."});
-    if(typeof thread_id !== "number" && typeof thread_id !== "string")
-      return callback({error: "Thread_id should be of type number or string and not " + typeof thread_id + "."});
 
-    if (typeof msg === "string") {
+    var msgType = utils.getType(msg);
+    var thread_idType = utils.getType(thread_id);
+
+    if(msgType !== "String" && msgType !== "Object")
+      return callback({error: "Message should be of type string or object and not " + thread_idType + "."});
+    if(thread_idType !== "Number" && thread_idType !== "String")
+      return callback({error: "Thread_id should be of type number or string and not " + thread_idType + "."});
+
+    if (msgType === "String") {
       msg = { body: msg }
+    } else if (msg.sticker) {
+      // Sticker can't be combined with body and/or attachment
+      delete msg.body;
+      delete msg.attachment;
     }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/sendMessage.js [hunk 1]
```diff
@@ -7,10 +7,15 @@ var log = require("npmlog");
 module.exports = function(mergeWithDefaults, api, ctx) {
   return function sendMessage(msg, thread_id, callback) {
     if(!callback) callback = function() {};
-    if(typeof msg !== "string") return callback({error: "Message should be of type string and not " + typeof msg + "."});
+    if(typeof msg !== "string" && typeof msg !== "object")
```
bsansouci: `typeof array` will also be object. The proper way to check for object would be
```js
Object.prototype.toString.call(msg) === '[object Object]'
```

I'll push a fix for the other places where we did this.

src/sendMessage.js
```diff
@@ -7,10 +7,15 @@ var log = require("npmlog");
...
+      return callback({error: "Message should be of type string or object and not " + typeof msg + "."});
```
bsansouci: Maybe same as above? `Object.prototype.toString.call(msg)`

Will return something like `[object Object]`, `[object String]`, `[object Array]`...

utils.js
```diff
@@ -53,6 +59,24 @@ function post(url, jar, form) {
   return request(op).then(function(res) {return res[0];});
 }
 
+function postFormData(url, jar, form, qs) {
```
bsansouci: Mmmh I'm not sure this is useful as a different function from `post`. Can we maybe add a 4th argument to `post`? @Schmavery thoughts?

bsansouci:
Wow thanks a lot for that, we've been wanting this for a while now! Great work. We'll merge as soon as all the little nitpicking is done :D

src/sendMessage.js
blazeu: It's seems like we are doing a lot of type checking everywhere. Shouldn't we use a library for it instead?

bsansouci: Well typeof works for strings, booleans and numbers. It just won't differentiate arrays from objects, and for arrays you have `Array.isArray(arr)`, so all we really need is the check for the object case. Maybe put in a helper function, not sure...

utils.js
Schmavery: If the only difference is including the `qs` in the form, I think it could be fine just added to `post`, yeah.

Schmavery:
This is awesome, love that you made it backward compatible too.  If you can make sure it supports `sendSticker`, we can deprecate that, I agree.

Am I correct in understanding that `api.uploadAttachment` isn't meant to be called by users, and you just exposed it in the api so that it could be in a separate file?

utils.js
blazeu: The ```Content-Type``` is multipart/form-data and it's using ```formData``` instead of ```form``` option, hence the function name.

blazeu:
Yes, but there might be a use case for this in the future.

utils.js
bsansouci: Ok that's fine then :)

blazeu:
I've added a helper function ```getType``` and ```checkType```. Based on http://toddmotto.com/understanding-javascript-types-and-reliable-type-checking/

```getType``` is a shortcut for ```Object.prototype.toString.call(obj).slice(8, -1) ```. The slice is used so that it will return ```Object``` instead of ```[object Object]```. also works for other types too.

```checkType``` is basically ```getType(obj) === type```.

**Comment by bsansouci** [comment]
Thanks for that, this looks good! I don't think `checkType` is all that usefull, in terms of number of characters that you gain from calling the function vs ===.
Now that we have this, could you use it everywhere instead of whatever we were doing before. (Don't worry about changing the rest of the code, just make sure your PR's up to date and we'll update the rest).

Also I agree with @Schmavery, we might as well support stickers in this. Technically `sendSticker` could be killed in favor of an all powerful `sendMessage` (might want to keep it but deprecate it). The only difference between `sendMessage` and `sendSticker` seems to be:
```js
  {
     // ....
     'message_batch[0][sticker_id]' : sticker_id,
      'message_batch[0][has_attachment]' : true,
      'message_batch[0][client_thread_id]' : "user:"+thread_id, // sendMessage doesn't even have this
}
```

blazeu:
I just tested it, it seems like all you have to do is change ```has_attachment``` to true and specify the ```sticker_id``` to send a sticker.

As for the ```checkType```, should i remove it and use ```getType(obj) === "type"``` instead ?

bsansouci:
Ok I trust you on that one :)
Yeah, I think the function `checkType` doesn't do much for us. The equality check is expressive/short enough.

blazeu:
Just do ```api.sendMessage({ sticker: sticker_id }, ...``` to send sticker.

Schmavery:
Hey @blazeu, thanks a lot for this, do you mind checking over the changes I made to the README and seeing if they match with what you wrote?
https://github.com/Schmavery/facebook-chat-api#sendMessage

blazeu:
You got it wrong on the basic message example, the field should be ```body``` instead of ```text```

bsansouci:
Fixed! Thank you so much :)

PR #58, opened 2015-08-09
Add url attachment

blazeu:
```
api.sendMessage({url: 'www.github.com/'}, ID);
```


index.js
```diff
@@ -68,7 +68,7 @@ function buildAPI(loginOptions, html, jar) {
 
   var apiFuncNames = [
     'listen',
-    'getUserID',
+    'getUserId',
```
bsansouci: Can you change that back to `ID`? We're trying to make everything consistent and decided on writing it `ID` instead of any other way.

index.js
```diff
@@ -80,13 +80,14 @@ function buildAPI(loginOptions, html, jar) {
     'removeUserFromGroup',
     'addUserToGroup',
     'sendTypingIndicator',
-    'getCurrentUserID',
+    'getCurrentUserId',
```
bsansouci: Same as above ^

src/getUri.js
```diff
@@ -0,0 +1,34 @@
+/*jslint node: true */
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+  return function getUri(uri, callback) {
+    if(!callback) callback = function() {};
```
bsansouci: Can you do [this](https://github.com/Schmavery/facebook-chat-api/blob/3dd39a2c0ae0b20c1202d02f4b70740789f28ccb/src/getUserId.js#L22) instead. It doesn't make much sense to not crash when you don't give a callback.

index.js
blazeu: whoops. sorry, when i was testing, the module didn't work because it cant load ```getUserID``` because you didn't change the filename.

src/sendMessage.js
```diff
@@ -80,10 +76,25 @@ module.exports = function(defaultFuncs, api, ctx) {
 
         send();
       });
+    } else if (msg.url || msg.uri) {
```
bsansouci: Can we avoid multiple ways of doing something? I'd rather just stick to `url` as it's what we use everywhere else.

bsansouci:
Just pushed the files with the right names. I'll merge as soon as you resolve the conflicts :)

blazeu:
But, I already changed the filename.

bsansouci:
Woops my bad. Thanks a lot :)

PR #64, opened 2015-08-29
customFunctions

garrylachman:
custom api function can be loaded via loginOptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ var login = require("facebook-chat-api");
 // Create simple echo bot
 login({email: "FB_EMAIL", password: "FB_PASSWORD"}, function callback (err, api) {
     if(err) return console.error(err);
-    
+
     api.listen(function callback(err, message) {
         api.sendMessage(message.body, message.thread_id);
     });
@@ -53,7 +53,7 @@ login({email: "FB_EMAIL", password: "FB_PASSWORD"}, function callback (err, api)
 <a name="login" />
 ### login(emailAndPassword, [options], callback)
 
-This function is returned by require(...) and is the main entry point to the API. 
+This function is returned by require(...) and is the main entry point to the API.
 
 Logs into facebook given the right credentials.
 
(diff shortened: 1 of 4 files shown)


Schmavery:
This is an interesting change, do you think you could give some justification for its necessity?

garrylachman:
first of all its great project.
because of easy login process i move some api calls like this one in my project (will commit the functions later, maby as addons - becouse its not related to chat)

https://www.facebook.com/ajax/typeahead/search/facebar/bootstrap/?viewer=534988564&__a=1&filter[0]=user

there is more api that can be added...

this change add me the ability to add my custom api functions that not part of the project...

Schmavery:
I'm starting to really like this idea.  @bsansouci can you see any downsides to this?  Could be good to include as part of v1.0.
The current implementation will override any functions with the same name, which seems fine to me.

garrylachman:
Its can expend the project to "underground" facebook api, not only chat api...

bsansouci:
I like the idea of making a more general API (higher than chat) that takes care of keeping a session on, and which would allow plugins. That said I don't think this should be developed under the master branch. Also this is a pretty hacky solution, but a good way to start. @garrylachman can you do this PR on the new dev-plugins branch?

garrylachman:
Yes... i will design it to be real addon solution and not hacky...


On Wed, Sep 2, 2015 at 5:27 AM, Benjamin San Souci <notifications@github.com
> wrote:

> I like the idea of making a more general API (higher than chat) that takes
> care of keeping a session on, and which would allow plugins. That said I
> don't think this should be developed under the master branch. Also this is
> a pretty hacky solution, but a good way to start. @garrylachman
> <https://github.com/garrylachman> can you do this PR on the new
> dev-plugins branch?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Schmavery/facebook-chat-api/pull/64#issuecomment-136916331>
> .
>



-- 
Best Regards,
Garry Lachman
Phone: +972 (0) 547759006
Skype: garryplastik
E-Mail: drgarry@gmail.com

Schmavery:
@garrylachman awesome, can you please submit an issue to track this feature where we can discuss design questions when they come up?

bsansouci:
@Schmavery a WIP PR would work too :)

Schmavery:
Right, feel free to just submit this code as a PR on the other branch before you go making any huge changes.  This is just a very design-focused change, so I want to make sure we stay up to date on our design decisions.

PR #67, opened 2015-09-02
added custom get request + tests

garrylachman:


bsansouci:
Hey @garrylachman, thanks for contributing! I'm not sure I see why we would need to go so low level. There has been some discussions about implementing a plugin system which you're welcome to contribute to (dev-plugins branch and PR #64).

garrylachman:
I think before addons implementation we need fist the ability to create some low level abstraction request functions, the next step is extend this function to create more customized request functions/addons and re-use the code...

bsansouci:
I see what you're saying but I don't think this should be exposed to the user. We already have internet get/post and ones with defaults for people making those plugins. But the user probably just wants to "use" the API.

bsansouci:
@Schmavery ^

garrylachman:
But something the user wants the ability to customise the request or make
some not included one...

On Sun, Sep 6, 2015 at 1:27 AM, Benjamin San Souci <notifications@github.com
> wrote:

> @Schmavery <https://github.com/Schmavery> ^
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Schmavery/facebook-chat-api/pull/67#issuecomment-138004388>
> .
>



-- 
Best Regards,
Garry Lachman
Phone: +972 (0) 547759006
Skype: garryplastik
E-Mail: drgarry@gmail.com

bsansouci:
Well that's why we'd have the plugin system :)
It's just that we've figured out most of the stuff for you already, so the user can built on top of what we've already built (like the defaultGet and defaultPost which include a bunch of data that a lot of requests need).

bsansouci:
I'm going to close this PR as we need to further discuss this. Can you start an issue if you're still interested in contributing?

PR #68, opened 2015-09-03
Get access token by app_id

davidyen1124:
Use Graph API Explorer api to get access token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,18 @@ __Arguments__
 * `thread_id` - Group chat ID.
 * `callback(err, end)` - A callback called when the query is done (either with an error or with null followed by a function `end` described above).
 
+---------------------------------------
+
+<a name="getAccessToken" />
+### api.getAccessToken(app_id, [callback])
+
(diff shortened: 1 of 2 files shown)


Schmavery:
Why is this useful to provide as part of a chat api?  Aren't there already packages dedicated to this?  I'm under the impression that this is exactly what https://github.com/jaredhanson/passport-facebook does.

davidyen1124:
We can use access token to [send page message](https://developers.facebook.com/docs/graph-api/reference/v2.4/page/conversations), I think it can be a feature in your project

Schmavery:
We already support sending messages as a page.  Are you suggesting that this should be implemented differently?

davidyen1124:
Yours is fine. Should I close this PR?

Schmavery:
Might as well.  Thanks for the contributing though, and let us know if you run into any problems with sending messages as a page.  This is a less-often used functionality to it's useful to get feedback. :smile:

Issue #56, opened 2015-08-04
Creating a Group Chat


Schmavery:
It seems we don't support creating a new group chat.  It might be possible to do it with `api.addUserToGroup`, but it's still not clear what to set as the `thread_id` and what the id of the new group chat will be.

JoseRoman:
The ID of chat group seem to be "Conversation-Thread_ID"

Schmavery:
This should be fixed by 7c88010 in v1.0

PR #69, opened 2015-09-05
Added ability to start a new group chat!

qcmaude:
This fixes Issue #56

### DIFF
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -114,7 +114,7 @@ module.exports = function(defaultFuncs, api, ctx) {
         for (var i  = 0; i < threadID.length; i++) {
           form['message_batch[0][specific_to_list][' + i + ']'] = "fbid:" + threadID[i];
         }
-        form['message_batch[0][specific_to_list][' + (threadID.length + 1) + ']'] = "fbid:" + ctx.userID;
+        form['message_batch[0][specific_to_list][' + (threadID.length) + ']'] = "fbid:" + ctx.userID;
         log.info("Sending message to multiple users: " + threadID);
       } else {
         form['message_batch[0][thread_fbid]'] = threadID;
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -18,7 +18,9 @@ module.exports = function(defaultFuncs, api, ctx) {
     if(msgType !== "String" && msgType !== "Object") {
       throw {error: "Message should be of type string or object and not " + threadIDType + "."};
     }
-    if(threadIDType !== "Number" && threadIDType !== "String") {
+
+    // Changing this to accomodate an array of users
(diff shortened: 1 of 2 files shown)


qcmaude:
It's in `sendMessage` because making a new group chat makes a call to the same endpoint (`send_message.php`) and uses very (very) similar form fields. The only big difference is that there is an added iteration through an array of user ids (and some checks put were in place to make sure that everything else remains the same).

src/sendMessage.js
```diff
@@ -103,52 +104,75 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+      // There are three cases here:
+      // 1. threadID is of type array, where we're starting a new group chat with users
+      //    specified in the array.
+      // 2. User is sending a message to a specific user.
+      // 3. No additional form params and the message goes to an existing group chat.
+      if(threadIDType === "Array") {
+        for (var i  = 0; i < threadID.length; i++) {
+          form['message_batch[0][specific_to_list][' + i + ']'] = "fbid:" + threadID[i];
         }
+        form['message_batch[0][specific_to_list][' + (threadID.length + 1) + ']'] = "fbid:" + ctx.userID;
```
bsansouci: Shouldn't this be `threadID.length` instead of `threadID.length + 1`?

src/sendMessage.js
```diff
@@ -103,52 +104,75 @@ module.exports = function(defaultFuncs, api, ctx) {
...
-      api.getUserInfo(threadID, function(err, res) {
-        if (err) {
-          return callback(err);
+    function sendContent(isSingleUser) {
+      // There are three cases here:
+      // 1. threadID is of type array, where we're starting a new group chat with users
+      //    specified in the array.
+      // 2. User is sending a message to a specific user.
+      // 3. No additional form params and the message goes to an existing group chat.
+      if(threadIDType === "Array") {
```
bsansouci: Could you use `isSingleUser` here?

src/sendMessage.js
```diff
@@ -103,52 +104,75 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+        .then(utils.parseAndCheckLogin)
+        .then(function(resData) {
+          if (!resData) {
+            throw {error: "Send message failed."};
+          }
+          if(resData.error) {
+            throw resData;
+          }
+
+          return callback(null, resData.payload.actions.reduce(function(p, v) {
```
bsansouci: This should only contain one thing right?

Schmavery: I think it's safer to do this, considering actions is an array.

src/sendMessage.js
qcmaude: LOLOL yep.

Schmavery:
I think this should be safe to merge.  Thanks for updating docs.

Issue #57, opened 2015-08-06
Logout


senhly:
I've tried to logout by clear cookie(set <b>ctx</b> to default value) but I don't think it is a good solution. I think we can logout through http://www.facebook.com/logout.php.
It seem facebook require parameter <b>fb_dtsg</b>,<b>h</b> and <b>ref</b> and then redirect to https://www.facebook.com/?stype=lo&jlou=...&smuh=....&lh=...
Does anyone know where is parameter <b>h</b> hiding? little help plz

Schmavery:
I agree that we should have this feature.  If you give us a couple days, we can probably find the time to look at it a little more closely.  We'll have to check exactly what information is being sent in the form before we can know how to emulate it.

senhly:
Cool ! now we can logout. :D
:( 
but the request after logout, facebook will ask for login to continue and next request will response html document so it cause error in <b>parseResponse</b>. any idea?

Schmavery:
@bsansouci now that you made the [change](f957e85) that all errors throw exceptions, can we simply throw an error "User Not Logged In" or something in `parseResponse`?

bsansouci:
@senhly what function does the request _after_ you've logged out?

@Schmavery though it would make sense in this case, I don't think it's a good error message. `parseReponse` makes the assumption that it receives JSON and it gets html, which can be caused by many things. 
We should be catching any error that parseResponse throws and calling the callback with it, so if you logout and do something you'll get an error. That seems about right to me.

Schmavery:
I think @senhly is talking about using other API functionality after logging out.  This should definitely cause an error, but it would be nice if it was a specific error and not just a parse problem.  I was assuming we would do some basic parsing of the html in `parseResponse` to verify that this was actually the problem.

bsansouci:
Mmh Ok, maybe we could make another `parseResponse` that will throw an error `{error: "Not logged in."}` that we use everywhere where we _are_ supposed to be logged in. The only place where we'd use the regular `parseReponse` is during the login process I'm guessing.

senhly:
@Schmavery was right, it cause error when call other function after logging out. By this way, I also notice that after logging out complete the pulling request in function <code>listen</code> still running. We could stop listen after log out. Am I right?

Schmavery:
Reopening so that we don't forget to fix this in listen.

PR #71, opened 2015-09-07
Fix logout

qcmaude:
Fixes issue #57 by not calling listen callback unless logged in :+1: :smile:

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,6 +54,7 @@ function buildAPI(globalOptions, html, jar) {
     jar: jar,
     clientID: clientID,
     globalOptions: globalOptions,
+    loggedIn: true,
     access_token: 'NONE'
   };
 
--- a/src/listen.js
+++ b/src/listen.js
@@ -35,7 +35,7 @@ module.exports = function(defaultFuncs, api, ctx) {
   var globalCallback = null;
 
   function listen() {
-    if(shouldStop) {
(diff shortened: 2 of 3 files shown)


Schmavery:
I'm wondering if, instead of `parseAndCheckLogin`, we should wrap every api function in something that checks login in index.

```
apiFuncNames.map(function(v) {
    var fn = require(".src/"+v)(defaultFuncs, api, ctx);
    api[v] = function(){
        if (ctx.loggedIn) fn.call(arguments);
        else throw new Error("Not logged in.");
    }
});
```

Schmavery:
^ @bsansouci When you get a chance, let me know what you think.  The downside is that it doesn't call the callback with the error, the upside is that you don't need to remember to check logged in status in every function and we make fewer unnecessary calls.  We could hack the callback problem by checking if the last argument is a function and calling it with the error ;)

bsansouci:
I like those changes. @schmavery, I actually don't think that is necessary. Any api calls will crash, and most will actually [catch the "not logged in" error](https://github.com/Schmavery/facebook-chat-api/blob/7c8801046c2770280c0e30bda758960d968920f9/utils.js#L363-L381). So I think a simple fix like this is good.

qcmaude:
LETS MERGE THIS :shipit:

PR #73, opened 2015-09-18
Docs update!

qcmaude:
If there are any ambiguities or typos, let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ For tests, create a `test-config.json` file that resembles `example-config.json`
 Pages can't start conversations with users directly; this is to prevent pages from spamming users.
 
 3. What do I do when `login` doesn't work?
-First check that you can login to Facebook using the website or app. If login approvals are enabled, you might be logging in incorrectly. For how to handle 2-factor auth, read our docs on [`login`](DOCS.md#login).
+First check that you can login to Facebook using the website or app. If login approvals are enabled, you might be logging in incorrectly. For how to handle login approvals, read our docs on [`login`](DOCS.md#login).
 
 
 You can find all of our documentation [here](DOCS.md).
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ For tests, create a `test-config.json` file that resembles `example-config.json`
 Pages can't start conversations with users directly; this is to prevent pages from spamming users.
 
 3. What do I do when `login` doesn't work?
-First check that you can login to Facebook using the website or app. If login approvals are enabled, you might be logging in incorrectly. For how to handle 2-factor authorization, read our docs on [`login`](DOCS.md#login).
+First check that you can login to Facebook using the website or app. If login approvals are enabled, you might be logging in incorrectly. For how to handle 2-factor auth, read our docs on [`login`](DOCS.md#login).
 
 
 You can find all of our documentation [here](DOCS.md).
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Facebook Chat API
+# Quick Start Guide
 The Official Facebook Chat API uses XMPP and is deprecated as of April 30th 2015. This is a non-official API that doesn't use XMPP.
 As of right now, the only way to automate the chat functionalities is to emulate the browser. This means doing the exact same GET/POST requests and tricking Facebook into thinking we're accessing the website normally. Because we're doing it this way, this API won't work with an auth token but requires the credentials of a Facebook account.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -108,46 +113,52 @@ __Review Recent Login__: Sometimes Facebook will ask you to review your recent l
...
+  + `participantNames`: An array containing only the first names of the other participants in the thread (sender included).
+  + `participantIDs`: An array containing the ids of everyone in the thread (sender included).
+  + `body`: The string corresponding to the message that was just received.
+  + `threadID`: The thread_id representing the thread in which the message was sent.
+  + `threadName`: The name of the receiver in a single person chat or the name of the group chat.
+  + `location`
+  + `messageID`: A string representing the message ID.
+  + `attachments`: An array of attachments to the message.
+  
+If `attachments` contains an object with type is `"sticker"`, the same object will contain the following fields: `url`, `stickerID`, `packID`, `frameCount`, `frameRate`, `framesPerRow`, `framesPerCol`, `spriteURI`, `spriteURI2x`, `height`, `width`, `caption`, `description`.
```
bsansouci: Would it be too lazy to just point at the `formatXXXX` function? We wouldn't have to update the docs, and it wouldn't be any less clear. (This applies to the paragraphs below too)

bsansouci:
@Schmavery mentioned something about having a section before the full docs that would contain maybe small examples or an FAQ. Something less brutal than just API References. What do you think?

README.md
```diff
@@ -263,20 +274,25 @@ Returns the currently logged-in user's Facebook user ID.
...
-* File/Image - Set field `attachment` to a readable stream or an array of readable streams.
+Various types of message can be sent:
+* *Regular:* set field `body` to the desired message as a string.
+* *Sticker:* set a field `sticker` to the desired sticker ID.
+* *File or image:* Set field `attachment` to a readable stream or an array of readable streams.
+* *URL:* set a field `url` to the desired URL.
+
+Note that a message can only be a regular message (which can be empty) and optionally one of the following: a sticker, an attachment or a url.
 
 __Tip__: to find your own ID, go to your own profile on Facebook and replace 'www' by 'graph' in the URL.
```
bsansouci: This actually doesn't work anymore :/. A replacement could be: go to your profile, click on your profile picture, look at the url that will look like `https://www.facebook.com/photo.php?fbid=10206056682155550&set=a.1652922442382.91979.|XXXXXXXXXX|&type=1&theater` where the `|XXXXXXXXXX|` are your `ID`.

bsansouci:
["will will"](https://github.com/Schmavery/facebook-chat-api/blame/0e508a014c11c304d970262e77a7f69974b4e739/README.md#L108) is a typo :p

README.md
```diff
@@ -263,20 +274,25 @@ Returns the currently logged-in user's Facebook user ID.
...
+### api.sendMessage(message, threadID, [callback])
 
 Sends the given message to the thread_id.
 
 __Arguments__
 
-* `message` - A string (for backward compatibility) or a message object as described below.
-* `thread_id` - A string, number, or array representing a thread. It happens to be someone's userId in the case of a one to one conversation or an array of userIds when starting a new group chat.
-* `callback(err, threadID)` - A callback called when sending the message is done (either with an error or with an confirmation object). `threadID` contains only the thread_id where the message was sent.
+* `message`: A string (for backward compatibility) or a message object as described below.
```
bsansouci: `(for backward compatibility)` ? Do we need to say this? Also this raises the question, do we really want to overload this function? @Schmavery

README.md
```diff
@@ -263,20 +274,25 @@ Returns the currently logged-in user's Facebook user ID.
...
+* `threadID`: A string, number, or array representing a thread. It happens to be someone's userId in the case of a one to one conversation or an array of userIds when starting a new group chat.
```
bsansouci: Can we just document as `string` or `array`.

bsansouci:
["`obj` is a mapping from userId to another object containing the following properties:"](https://github.com/Schmavery/facebook-chat-api/blame/0e508a014c11c304d970262e77a7f69974b4e739/README.md#L377) Related to the comment above, can we just point to the format function?

bsansouci:
https://github.com/Schmavery/facebook-chat-api/pull/73/files#diff-04c6e90faac2675aa89e2176d2eec7d8R472 Can we just say `sends a typing indicator`? Because it'll be different depending on the platforms.

README.md
Schmavery: I really don't want to do this.

README.md
bsansouci: Actually just open the inspector, look at the cookies, the `userID` should be `c_user`.

README.md
Schmavery: I think we can still overload the function and remove the (for backward compatibility)

bsansouci:
Ok I fixed the small nitpicking in e847462.

Schmavery:
I think all that's left is the Quick Start Guide.
The README should be renamed to DOCS.md or something and a new README should be created containing install instructions, how to do basic listen/sendMessage, and a couple FAQs (This could contain things like how to run the tests, why sendMessage sometimes doesn't work as a page, troubleshooting tips when login doesn't work etc)

bsansouci:
I'd agree with that ^

qcmaude:
@bsansouci But 20 factor authorization is the future.

bsansouci:
squashhhhh <3

bsansouci:
Thankkkks :dancer:

PR #76, opened 2015-09-24
getMessages (retrieve list of messages from thread)

bbatliner:
I'm creating an app for Facebook messages and needed to retrieve the existing messages from a thread. I copied "getThreadList.js" and modified it for "thread_info.php". I don't know what errors this endpoint will return, so I left the error checking as-is. Please make sure this meets your standards and I would love to see it added!

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,6 +130,7 @@ function _login(email, password, loginOptions, callback) {
         'sendDirectSticker',
         'sendSticker',
         'setTitle',
+        'getMessages',
         'getThreadList',
         'markAsRead',
         'sendMessage',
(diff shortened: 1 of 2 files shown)


Schmavery:
Thanks for the contribution!  It actually looks like we just pushed exactly what you need a couple days ago to the pre-release-v1.0 branch just a couple days ago.  Does `getThreadHistory` do what you need it to do?

bbatliner:
Hm, yes, `getThreadHistory` is doing the same thing. Thank you for pointing that out! I'm looking forward to the 1.0 release :) I'm going to go ahead and close this request, since 1.0 will include it.

Schmavery:
Awesome, I hope that works out for you, let us know if you have any problems.

PR #78, opened 2015-09-26
Limit thread query from 1 --> 21

bbatliner:
See the following image:
image
The Facebook API uses limit: 21 when it hits search_threads.php. Unless it was intentional to only search for a single thread, I think it was a typo.


Schmavery:
This is because we originally only returned the first element, but I agree that this makes more sense.

PR #79, opened 2015-09-27
Add "video" to attachment list

bbatliner:
This switch statement didn't account for "video" attachments. See this gist for a full video object returned from the Facebook API: https://gist.github.com/bbatliner/c5c702b9dc9c07c080c3. Let me know if you need/don't need any of the properties.


Schmavery:
Thanks, @bsansouci just added b177725 to match this on the sendMessage side.

PR #80, opened 2015-09-29
Account for undefined attachments in formatAttachment

bbatliner:
I was trying to load some old conversations and received an error saying could not access `map` of `undefined`. I think Facebook used to pass an undefined array in `attachments`, and this small change accounts for that.


bsansouci:
Nice catch! Thanks :dancer:

PR #81, opened 2015-10-02
Return timestamp in sendMessage

bbatliner:
Send back Facebook's message timestamp in the sendMessage endpoint.

I think this is necessary because Facebook's timestamps are different than my local machines'. For example, when I receive a message (with `listen()`), my Date.now() is 15 seconds off from the message's timestamp that I received from Facebook. This causes issues when users send/receive messages in a short amount of time. Using Facebook's timestamps everywhere should resolve the timestamp discrepancies I'm seeing.

Side note: is there a reason the whole message object isn't sent back?

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -432,7 +432,7 @@ __Arguments__
 
 * `message`: A string (for backward compatibility) or a message object as described below.
 * `threadID`: A string, number, or array representing a thread. It happens to be someone's userId in the case of a one to one conversation or an array of userIds when starting a new group chat.
-* `callback(err, messageInfo)`: A callback called when sending the message is done (either with an error or with an confirmation object). `messageInfo` contains the `threadID` where the message was sent and a `messageID`.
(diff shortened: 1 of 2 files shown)


Schmavery:
I think including the timestamp is a good idea.  Right now you can get around this problem with selfListen but that's kinda gross.  I'm generally against cluttering up responses, but I'm definitely open to arguments for including more fields.  I already feel like we're returning too much for some api calls.  Is there any other data you think could potentially be useful?

Schmavery:
@bbatliner one request: can you update the docs to reflect this change?  Then I'd be happy to merge it :)

bbatliner:
Done!

None of the other information in the message model is particularly useful. The application calling `sendMessage` should know the message it sent, any attachments, etc. The human-readable timestamps `timestampAbsolute`, `timestampRelative`, and `timestampDatetime` could be useful, but they can be calculated by the application if need be.

Schmavery:
Thanks!  So hard to keep those docs up to date ;)

bbatliner:
If I get the time I'll try documenting some of the functions not on the
DOCS page. Good project you've made! Thanks for being so cooperative :)

PR #95, opened 2015-11-10
use ctx to get userID when requesting friend list

RandallKent:


bsansouci:
Oh whoops. Good call!

PR #106, opened 2015-11-27
only "from", "st" and "type" is received in typing indicator event

nripendra:
Not sure if this would be the correct fix. But I was having issue due to this.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -290,10 +290,10 @@ function formatEvent(m) {
 function formatTyp(event) {
   return {
     isTyping: !!event.st,
-    from: event.from.toString(),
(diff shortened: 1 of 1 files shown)


Schmavery:
Can you share what data is actually received in the event?  It seems like we should probably be getting more than `event.st` and `event.from_mobile`.  It doesn't really make sense to return empty string as a threadID, for instance.

nripendra:
Sorry, can't seem to reproduce this issue right now. I'll try to post data as soon as I get that issue again.

bsansouci:
Ok, feel free to reopen when you're able to reproduce it.

PR #108, opened 2015-12-06
Changed to threadID

lasseborly:
Changed the event.thread_id to event.threadID so that it corresponds with the rest of the documentation.


bsansouci:
Wow good call, I don't know how we missed that @Schmavery. Thanks!

Schmavery:
:+1:

PR #111, opened 2015-12-06
Fix internal link to api.getThreadHistory

jobbogamer:
The link in the docs currently jumps to searchForThread.


bsansouci:
Thanks!

PR #113, opened 2015-12-12
Fix type inconsistency in read_receipt event

blazeu:
``reader`` and ``threadID`` in all events is always string, except for read_receipt.

Also, i want to ask, why is the key which contain the sender id is different in every event? (senderID, author, from, reader)

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -300,9 +300,9 @@ function formatTyp(event) {
 
 function formatReadReceipt(event) {
   return {
-    reader: event.reader,
+    reader: event.reader.toString(),
     time: event.time,
-    threadID: event.realtime_viewer_fbid,
+    threadID: event.realtime_viewer_fbid.toString(),
     type: 'read_receipt',
   };
 }


Schmavery:
Thanks for catching this.
The effective reason that the key is different is because the object that we get from facebook is frequently uses different keys.  Also these keys tend to more closely match the semantics of the event.  That said, I wouldn't be against bringing some consistency to these events now that we have a number of them and releasing a v1.1.0 or something like that.
If we were going to do that, I'd want to update the docs to have a separate section describing events instead of having everything lumped under `api.listen`, since that section is getting pretty crowded as we add more and more events.  Thoughts @bsansouci?

bsansouci:
I'm a little scared to set such a "standard" because what happens when something doesn't fit that? If we set the expectation that _every_ event object has a senderID we need to make sure it's perfectly consistent. Does `senderID` make sense for an event like "X left the thread" or even "X was added by Y" then who's the "sender"?
On top of that potential inconsistency, why would we want to set such an expectation for events? I can't imagine a reason why you'd want to handle every event the same way. If anything I see reasons why we'd want to discourage that. Each event has different semantics, as you said @Schmavery, so by adding consistency we'd effectively kill those semantics from the object structure and have them implied through documentation which can go out of sync, or never read by a user (cf: all of the problems with having semantics enforced in docs).

Unless you can find a compelling argument @blazeu @Schmavery, I'd say we shouldn't do that.

bsansouci:
Thanks for this PR in any case :)

PR #115, opened 2015-12-16
Fix: change pageId to pageID on DOCS.md

leonardoobaptistaa:
If login function is used with {pageId: 'xxx'}, it raises a Warning:
WARN Unrecognized option given to setOptions pageId.

Changing to pageID fixes the problem


Schmavery:
Thanks!

PR #117, opened 2015-12-21
removed duplicate index

rohitguptadev:
There was a login issue because of Syntax error (Duplicate data property in object literal not allowed in strict mode). So I  removed the Duplicate index in src/getFriendsList.js

### DIFF
--- a/src/getFriendsList.js
+++ b/src/getFriendsList.js
@@ -12,13 +12,13 @@ var GENDERS = {
   3: 'female_singular_guess',
   4: 'male_singular_guess',
   5: 'mixed_singular',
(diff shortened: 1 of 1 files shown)


Schmavery:
Thanks for spotting this.  This is actually the wrong solution because this is an enum that is copied from facebook's code (which contains duplicate keys), so we can't just bump the numbers without losing the meaning here.
The correct solution would be to replace 
```
    5: 'mixed_singular',
    5: 'mixed_plural',
```
with
```
    5: 'mixed',
```
or something similar.

Schmavery:
Fixed by #120.

PR #120, opened 2015-12-24
Fix duplicate property in gender list

Weetbix:
Fixes ES5 strict mode error. 

This seems to be allowed in ES6 strict but Genders[5] just gets reassigned to mixed_plural. I think "mixed" is a reasonable fix given it looks like it may be either mixed singular or plural.


Schmavery:
Thanks.

PR #121, opened 2015-12-24
Correct errors in documentation

Weetbix:
Fixes a few errors in the API documentation

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -135,7 +135,7 @@ var login = require("facebook-chat-api");
 login({email: "FB_EMAIL", password: "FB_PASSWORD"}, function callback (err, api) {
     if(err) return console.error(err);
 
-    api.changeArchivedStatus(123456789, true, function callback(err) {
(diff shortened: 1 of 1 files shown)


Schmavery:
Thanks for catching these.
In the changeArchiveStatus docs, can you change all mentions of "state" to the use the word "status"?

On line 126, can you make it say "The id(s) of the threads you wish to archive/unarchive."
On 127, please specify that the value should be a boolean.

If you could add a short example, that would also be awesome.

Weetbix:
No problem, should be good now. I also added a description of what archiving is.

This is a great project, so thanks for all your work and merry christmas! :santa:

Schmavery:
I would maybe change your thread ID to 0000000000000 to stay consistent with the other examples.  Other than that looks awesome and I'm always happy to merge doc fixes :D

Weetbix:
Done :+1:

Schmavery:
Weird, thought I already merged this.  Thanks again.

PR #131, opened 2016-01-06
It seems 2FA (login approvals) do not work any more

mehcode:
They look like they used to work but Facebook has changed the format of the form.

I spent some time on this (out of time for today) and I found one tweak we need. :beers: if you could take it from here.


Schmavery:
Thanks for catching this, happens a lot.
I think we should be trying to transition to a method of checking that doesn't rely on your language being set to English if possible.

PR #134, opened 2016-01-06
:bug: Asking for message history from a thread with no message history

mehcode:


bsansouci:
Nice catch! Thanks :)

PR #140, opened 2016-01-10
fix thread history docs

dionyziz:
Improve the docs of the `getThreadHistory()` function.

* Fix wrong link in `README`
* Clarify that callback of `getThreadHistory` takes second `history` parameter


Schmavery:
Thanks a lot @dionyziz, not sure how we missed that.

PR #155, opened 2016-01-20
Fix message type error log

enhdless:
The error message for an invalid message type was logging the wrong variable.


Schmavery:
Thanks.

PR #162, opened 2016-01-25
Changed POST from GET in request

GAMELASTER:
Hmm, interesting (if it works) how can working this function, but in Browser is request sent as POST, not GET.


Schmavery:
Sorry, I don't understand why you're doing this.  Is there a related issue that it fixes?

GAMELASTER:
image
The request is POST, not GET, so it is interesting how can GetUserInfo works until this :-D

bsansouci:
Wow nice catch. Thanks :)

Issue #164, opened 2016-01-26
Package Breaking, JS Object Error


mjkaufer:
Error on line 194 of `utils.js`, https://github.com/Schmavery/facebook-chat-api/blob/master/utils.js#L194

I'll fix this

mjkaufer:
Also unrelated, I think message types are not working property, but that's not as bad.

mjkaufer:
Check out https://github.com/Schmavery/facebook-chat-api/pull/165

bsansouci:
I responded on the PR. Our package specifies that you need nodejs ≥ 4.x, which I _think_ is when they started supporting this. Please correct me if I'm wrong.

mjkaufer:
Sorry - I must have missed that

mjkaufer:
Yeah, I was running 0.12.2, which would explain the issue - everything works wonderfully now

bsansouci:
<3

PR #165, opened 2016-01-26
Fixed bug in generatePresence method

mjkaufer:
See https://github.com/Schmavery/facebook-chat-api/issues/164

Improper JSON formatting

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -177,7 +177,8 @@ function presenceDecode(str) {
 
 function generatePresence(userID) {
   var time = Date.now();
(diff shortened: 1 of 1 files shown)


bsansouci:
Hey @mjkaufer we used computed properties on purpose here. See https://github.com/lukehoban/es6features#enhanced-object-literals. This is part of newer versions of nodejs https://nodejs.org/en/docs/es6/#which-features-ship-with-node-js-by-default-no-runtime-flag-required

mjkaufer:
Oh whoopsies, my bad - I guess I have an outdated version of node

PR #171, opened 2016-02-05
If the user has no threads at all, this is undefined.

mehcode:


bsansouci:
Good catch. Thanks :)

PR #183, opened 2016-03-01
fixed deleting body on sticker presence

doubiz:
The body doesn't get deleted when a message has a sticker.
Deleted body from `form` object rather than `msg` object.

### DIFF
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -142,8 +142,7 @@ module.exports = function(defaultFuncs, api, ctx) {
       form['message_batch[0][sticker_id]'] = msg.sticker;
 
       // Sticker can't be combined with body
-      delete msg.body;
-
+      delete form['message_batch[0][body]']
       send();
     } else {
       send();


bsansouci:
Nice catch! Thanks for that :)

Schmavery:
What was the behavior before this change.  Did the call receive an error when sending a sticker with a body?

doubiz:
@bsansouci your welcome, glad to help :)
@Schmavery No it didn't receive any errors while sending with a body. 
The problem occurs on the mobile app it only shows the message body, it doesn't show the sticker. However in the browser it shows the body along with the sticker.

Schmavery:
Thanks for clarifying, @doubiz.  I love the little inconsistencies between clients ;P

I think this might become a design question as to whether we want to explicitly allow of these kinds of behaviours or not in the future, so don't take it personally if this change gets reverted at some point.

We already allow you to do some things that the client doesn't really let you do, like sending messages with empty string as a body (which also behaves differently on different clients).  Personally I'm in favour of allowing people to do more things when it doesn't cause an error, but it's something that can be discussed in the future.

PR #190, opened 2016-03-03
Add read event

HowardLoTW:
This feature is able to know what time does the user himself read someone's messages.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -381,12 +381,12 @@ If `type` is `"typ"` then the object will have the following fields:
 - `threadID`: Current threadID.
 - `from_mobile`: Boolean representing whether or not the person's using a mobile device to type.
 
-If `type` is `"read_receipt"` then the object will have the following feilds:
+If `type` is `"read_receipt"` then the object will have the following fields:
 - `reader`: ID of the user who just read the message.
 - `time`: The time at which the reader read the message.
 - `threadID`: The thread in which the message was read.
 
-If `type` is `"read"` then the object will have the following feilds:
+If `type` is `"read"` then the object will have the following fields:
 - `threadID`: The threadID representing the thread in which the message was sent.
 - `time`: The time at which the user read the message.
 
--- a/DOCS.md
+++ b/DOCS.md
@@ -388,7 +388,6 @@ If `type` is `"read_receipt"` then the object will have the following feilds:
 
 If `type` is `"read"` then the object will have the following feilds:
 - `threadID`: The threadID representing the thread in which the message was sent.
-- `userID`: The user ID.
 - `time`: The time at which the user read the message.
 
 Difference between `"read_receipt"` and `"read"`:
--- a/DOCS.md
+++ b/DOCS.md
@@ -379,7 +379,7 @@ If `type` is `"typ"` then the object will have the following fields:
 - `isTyping`: Boolean representing whether or not a person started typing.
 - `from`: ID of the user who started/stopped typing.
 - `threadID`: Current threadID.
-- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type.
+- `fromMobile`: Boolean representing whether or not the person's using a mobile device to type.
 
 If `type` is `"read_receipt"` then the object will have the following fields:
 - `reader`: ID of the user who just read the message.
--- a/DOCS.md
+++ b/DOCS.md
@@ -376,15 +376,25 @@ If `type` is `"event"` then the object will also have those fields:
 - `author`: The person who performed the event.
 
 If `type` is `"typ"` then the object will have the following fields:
-- `isTyping`: Boolean representing whether or not a person started typing
-- `from`: ID of the user who started/stopped typing
-- `threadID`: Current threadID
-- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type
+- `isTyping`: Boolean representing whether or not a person started typing.
+- `from`: ID of the user who started/stopped typing.
+- `threadID`: Current threadID.
+- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type.
 
-If `type` is `"read_receipt"` then the object will have the following fileds:
-- `reader`: ID of the user who just read the message
-- `time`: the time at which the reader read the message
-- `threadID`: the thread in which the message was read
+If `type` is `"read_receipt"` then the object will have the following feilds:
+- `reader`: ID of the user who just read the message.
+- `time`: The time at which the reader read the message.
+- `threadID`: The thread in which the message was read.
+
+If `type` is `"read"` then the object will have the following feilds:
+- `threadID`: The threadID representing the thread in which the message was sent.
+- `userID`: The user ID.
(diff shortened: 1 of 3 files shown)


src/listen.js
```diff
@@ -45,6 +45,9 @@ module.exports = function(defaultFuncs, api, ctx) {
       case 'read_receipt':
         globalCallback(null, utils.formatReadReceipt(event));
         return true;
+      case 'read':
```
Schmavery: Can you add a comment explaining why this is different than a read receipt?

utils.js
```diff
@@ -399,6 +399,20 @@ function formatReadReceipt(event) {
   };
 }
 
+function formatRead(event) {
+  return {
+    tids: event.tids,
```
Schmavery: The key names on this object don't follow the style of other objects.

Schmavery:
I left some comments.  It's important for the event object go be consistent with other events.  There should also be a new entry in the docs for this event.

utils.js
HowardLoTW: Hello, so what key name should I change from ```tids``` ? ```threadIDs``` ?
the "tids" looks like => tids: [ 'id.222346744472064' ]
I'm not sure whether it is thread IDs or not.
All the keys here are simply as same as Facebook's ```event``` object.
It would be nice that you can give me some suggestions for the naming of keys .
Thanks.

src/listen.js
HowardLoTW: I would add a comment like this: 
```// read event triggers when the user read other people's messages ```
Is this command OK ?

Thanks.

**Comment by HowardLoTW** [comment]
Hello, I also left some questions for you.
If it's OK, I will update the code and docs then.

Thanks.

File: src/listen.js (hunk 1 above)
Comment (Schmavery) [inline src/listen.js]: Yes that sounds fine.  Please add the same note to your documentation when you add that.

File: utils.js (hunk 2 above)
Comment (Schmavery) [inline utils.js]: If you look at the other objects, you will see they are all camelcased (`folderInfo` instead of folder_info).  We try to do some work to gain consistency as well by combining together various properties.  I would probably suggest something like the following:
```js
{
    threadID: event.chat_ids[0] || event.thread_fbids[0], // ID of the thread you read
    timestamp: event.timestamp,
    userID: event.realtime_viewer_fbid, // Your ID
    type: 'read'
}
```
Arguably the userID could be removed if we know that it will always be your id.  If `mark_as_read` is always true we shouldn't include it.  I doubt that `folder_info` will be useful to anyone.  As a note, we treat a user's Facebook userID as their threadID in a single-user chat with them (that explains the `threadID` property).
@bsansouci anything I'm missing?

src/listen.js
HowardLoTW: Ok. No problem here.
Let's confirm there is any problem about the key naming issues.
If there's no problem, I'll give you a pull request.
Thanks.

Schmavery: Just in case you didn't know, you should be able to update this pull request with the new commits rather than creating a whole new one.

bsansouci:
Yup looks all good! Thanks a ton @HowardLoTW :D

HowardLoTW:
@bsansouci @Schmavery 
Hello ! I also found that the ```threadID``` in  ```read_receipt``` event is probably wrong.
The original source code : 
```
function formatReadReceipt(event) {
  return {
    reader: event.reader.toString(),
    time: event.time,
    threadID: event.realtime_viewer_fbid.toString(), // <------ This is actually the other user ID, not the thread ID.
    type: 'read_receipt',
  };
}
``` 
Want me to fix this in another pull request by the way ?

Schmavery:
Ok, good catch, please replace it with `thread_fbid`
```js
function formatReadReceipt(event) {
  return {
    reader: event.reader.toString(),
    time: event.time,
    threadID: event.thread_fbid.toString(),
    type: 'read_receipt',
  };
}
```

HowardLoTW:
@bsansouci  chat_ids here is not a thread ID, it contains user ID.

bsansouci:
We use the `userID` as a `threadID` when you're in a single user chat.

utils.js
```diff
@@ -399,6 +399,16 @@ function formatReadReceipt(event) {
   };
 }
 
+function formatRead(event) {
+  return {
+    threadID: event.tids[0],
+    userID: event.realtime_viewer_fbid.toString(),
+    senderID: other_user_fbids[0].toString(), || thread_fbids[0].toString(), || '', // senderID will be an empty string if the user is chatting in a group.
```
bsansouci: You have too many commas here :p

HowardLoTW: Sorry, my mistake.
I'll fix that.

HowardLoTW:
@bsansouci 
After merging this pull request, I'll then create another pull request to fix the threadID in formatReadReceipt(): 
```
function formatReadReceipt(event) {
  return {
    reader: event.reader.toString(),
    time: event.time,
    threadID: (event.thread_fbid) ? event.thread_fbid.toString() : event.realtime_viewer_fbid.toString(),
    type: 'read_receipt',
  };
}
```
because it seems that I could only create another pull request after merging this current pull request.

HowardLoTW:
@bsansouci  I think what you mean is `userID` is always my ID, not `senderID`.
Since `senderID` is the same as `threadID`, I've remove `senderID` in this commit.
So do you want me to remove `userID`, too ? (I just saw your comment after I commited it)

bsansouci:
Yes I think you don't need to be reminded of your own ID :p

DOCS.md
```diff
@@ -376,15 +376,23 @@ If `type` is `"event"` then the object will also have those fields:
...
-If `type` is `"read_receipt"` then the object will have the following fileds:
-- `reader`: ID of the user who just read the message
-- `time`: the time at which the reader read the message
-- `threadID`: the thread in which the message was read
+- `isTyping`: Boolean representing whether or not a person started typing.
+- `from`: ID of the user who started/stopped typing.
+- `threadID`: Current threadID.
+- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type.
+
+If `type` is `"read_receipt"` then the object will have the following feilds:
```
bsansouci: please change to `fields`

DOCS.md
```diff
@@ -376,15 +376,23 @@ If `type` is `"event"` then the object will also have those fields:
...
-- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type
-
-If `type` is `"read_receipt"` then the object will have the following fileds:
-- `reader`: ID of the user who just read the message
-- `time`: the time at which the reader read the message
-- `threadID`: the thread in which the message was read
+- `isTyping`: Boolean representing whether or not a person started typing.
+- `from`: ID of the user who started/stopped typing.
+- `threadID`: Current threadID.
+- `from_mobile`: Boolean representing whether or not the person's using a mobile device to type.
```
bsansouci: We just updated `from_mobile` to `fromMobile`, can you make the same change so we can merge :)

HowardLoTW: Sure.

bsansouci:
Thanks a lot @HowardLoTW  :D

HowardLoTW:
Hello, will it be updated to npm package ?
Thanks :-)

PR #193, opened 2016-03-06
New functions: accept/ignored message requests

abdawoud:


Schmavery:
Can you explain what a request is?  As in what is the value we are passing to this function?  It doesn't seem like there's a way to get this within the api, is there?  Do we need to add a "message_request" event?
Can we merge these two functions together under a name like "handleMessageRequest(accept)" that takes true when you want to accept and false for ignore?
This is awesome otherwise, though ideally we would have some docs as well.

GAMELASTER:
@Schmavery When some man trying to contact you and its not a friend of your friend or you are not in same group, message is sent to "Others" thread, where you can "accept" or "ignore" a request to chat, when you accept, the message and whole thread is moved into normal group, else its ignored (of course)

Schmavery:
I'm sorry, I should have been clearer.  I mean in the context of the API.  I understand what a message request is, but I don't understand what value we're passing into these functions.

Schmavery:
@abdawoud ping...

bsansouci:
Hey @abdawoud maybe you should collaborate with @how2945ard since he's working on https://github.com/Schmavery/facebook-chat-api/pull/247

Schmavery:
I think this functionality is merged now.  Thanks a lot for your investigation of this feature @abdawoud !

PR #196, opened 2016-03-11
Make revision to parse only once

GAMELASTER:
I think its useless to parse everytime a same revision number, when the html variable is still same too.


Schmavery:
Is this a performance problem for you?

PR #199, opened 2016-03-13
Patch bug of non working messenger scripts when logining from AppState

GAMELASTER:
Okay, so in Messenger.com works same cookies in Facebook.com .
Change color and emoji dont work for me because when is user logged via AppState, in AppState isnt a Messenger.com cookies and its not added after login, so I added that!

But I dont think this is a good idea to fix this problem! I think it will be better to roll back the changes of saving a messenger.com cookies, and only create function "makeMessengerCookies" for scripts what requesting to messenger.com.

### DIFF
--- a/index.js
+++ b/index.js
@@ -273,6 +273,11 @@ function loginHelper(appState, email, password, globalOptions, callback) {
     appState.map(function(c) {
       var str = c.key + "=" + c.value + "; expires=" + c.expires + "; domain=" + c.domain + "; path=" + c.path + ";";
       jar.setCookie(str, "http://" + c.domain);
+			if(c.domain.indexOf("facebook.com") > -1)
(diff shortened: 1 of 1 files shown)


Schmavery:
I'm not sure if I understand.  If getAppState doesn't return messenger cookies, that will be a bug.  If you're using an appState from before the change, we can't help with that.
Basically, the regular appState should include all required cookies.  We won't implement a "makeMessengerCookies".

Schmavery:
Ahh, yes, I just checked, and the problem is actually that getAppState doesn't output the messenger.com cookies, as you can see here: https://github.com/Schmavery/facebook-chat-api/blob/924f3532e027896b14e35754f495bc45c1d70e28/index.js#L63-L66

Issue #161, opened 2016-01-25
Hit the right server


Schmavery:
We currently only ever hit 0-edge-chat.facebook.com.
This could be causing a couple of issues, like #156 and #158.

mehcode:
Going to wager a guess that the problems are actually the static usage of `seq` and `partition`:

https://github.com/Schmavery/facebook-chat-api/blob/master/src/listen.js#L24-L25

As those parameters smell to be used for something _important_. `*-edge-chat` is a static CNAME to `star.c10r.facebook.com` which is a DNS-based load / region balancer. The specific `*-edge-chat` you use has no bearing on the IP that is resolved. The client seems to pull one out of a hat (using Math.random) and then use local storage so it doesn't change for at least ~1.6 minutes (I haven't been looking at this long enough to see the at most time).

mehcode:
#156 looks like facebook chat was simply offline or that the local DNS resolution failed. As `getaddrinfo` failed to retrieve address information for the CNAME.

#158 could mean facebook was having issues. I don't see enough testing to justify a problem with the library (and not a blip in facebook's network).

---

Looking through those issues I'm thinking there isn't any real "problem" with just using 0-edge-chat but as the client does cycle through `seq` and `partition` I assume they are needed for something or other.

Schmavery:
Thanks for your investigation @mehcode 
We haven't taken the time to investigate partition (afaik @bsansouci) because it's always been -2 when we've seen it.
Maybe switching to another *-edge-chat when a call fails?  Not sure, but it would be nice to be able to handle #156.  If we can find out how the client figures out the range of numbers to choose from, that would be cool.

mehcode:
>  If we can find out how the client figures out the range of numbers to choose from, that would be cool.

```
Math.floor(Math.random() * MAX_CONNECTIONS) % 7
```

`max_conn` is from the reconnect call which is consistently `6` over here.

That number (from 0 to 6) is then stored in local storage along with a timestamp to make sure you stay on that channel for at least `100 * 1000` milliseconds if you refresh. I didn't leave the inspector open long enough (I did for a couple hours) to see if it switched on its own.

Schmavery:
[...]

PR #206, opened 2016-03-17
Hitting the right server

GAMELASTER:
I investigated how is *-edge-chat generated, if it is a really random or not.
The result, its really random!
For explain I write the process lower, but first to code.
I written generateSubdomain similar to original code for future change (max_conn, or L can be changed). I supposed its a best way, if not, sorry!

So, this is explain:
The function of generating is in ChannelSubdomain (for search in Chrome Debugger, use CTRL SHIFT F for search in Sources tab), the Q function (exported as allocate), with three parameters, only first is used, and its max_conn (search this for executing append).
In allocate (returning a U % L) function is first created an U variable, what I used in my function. Then is used random to get an S, random generated number from U. And N, N has same value as S, all of the time (if code failure, N is set to S). In block with N is something like checking of something with IDs, I really have not idea what it is, but its not doing anything with final version, a random U, and after is set some values for channel pubs from N. So, its random!
image


Schmavery:
In the browser, is the subdomain different every time you pull, or does it stay the same for a while?
I thought it usually stays the same, but maybe that doesn't matter too much.

One issue I see is that all your values are hardcoded constants.  In fact, your whole generateSubdomain function is roughly equivalent to `Math.floor(Math.random() * 7)`...
Ideally we should be able to get these values from somewhere, unless we can establish that they're also hardcoded in the client, in which case, we might be able to get away with it.

Schmavery:
As far as I'm concerned, a fix for this would also include a strategy for caching the subdomain and retrying with a new subdomain if the call fails.

GAMELASTER:
Subdomain is changed every refresh of website. I Will look if it's not changed more times while a runtime

mehcode:
> Subdomain is changed every refresh of website. 

Yes and no. On a refresh facebook will try to keep you at the same subdomain unless its been more than `100 * 1000` milliseconds since the refresh. 

I did a small bit of investigation myself on this: https://github.com/Schmavery/facebook-chat-api/issues/161

We should probably just implement something that changes the subdomain every so often and re-authorizes there. 

`Math.floor(Math.random() * 7)` is more than enough. 

The facebook client code is just plain weird.

GAMELASTER:
@mehcode, yep, you are right! Yes, FB code is creepy.

Schmavery:
My issue is mainly with hardcoding a '7' into our code empirically.  I'm not saying we don't do that in other places, but it's not ideal.
Not planning on merging this unless `generateSubdomain()` is cleaned up and the caching is addressed.  Thanks for investigating further into that @mehcode.

Honestly, with them picking servers randomly rather than doing any kind of active load-balancing or anything, I feel this is a low priority change that is unlikely to fix anything unless we are able to change servers upon failure of the request.

GAMELASTER:
Well, I can close this pull request.

(omg, sorry for broken things with pull request)

PR #207, opened 2016-03-18
Adds basic implementation of changing group chat image

diceroll123:
Didn't make it TOO shiny as it is rather late, but here's a working barebones version. No error checking really... It's also slightly sloppy in the sense that I haven't used promises yet on my own and thus used bluebird incorrectly, pushing a single upload into an array and processing the array. I just repurposed the upload function from sendMessage. But it works! Feel free to pretty it up. :smile:

Usage:

```
api.changeGroupImage("./avatar.png", threadID, function callback(err, message) {
    if(err) return console.error(err);
});
```


GAMELASTER:
Tested, works nice! I think the code is okay! Really nice work!

Schmavery:
Thanks for contributing.
Why doesn't this match the signature of `sendMessage` when it comes to uploading an image?  `sendMessage` accepts a stream and this takes a string.

This should also add an entry to the docs (ideally a test as well).

diceroll123:
In the interest of full disclosure, testing kept being buggy for me, so I was unable to test properly using `npm test`.

Here's what _should_ work, but of course it may be incorrect.

```js
it('should change group chat image', function (done){
  listen(done, function (msg) {
    return msg.type === 'message' && msg.body === body;
  });
  api.changeGroupImage(fs.createReadStream("test/data/test.png"), groupChatID, function(err, info){
    checkErr(done)(err);
    assert(groupChatID === info.threadID);
  });
});```

I've done all I can do. :confused:

Schmavery:
Understandable, our tests are completely undocumented and more than a little strange.  Thanks for trying.  Docs look great.  @bsansouci everything look fine to you?

bsansouci:
LGTM!

PR #211, opened 2016-03-21
Remove extraneous and invalid main

AdamBurgess:
In node, `require('facebook-chat-api')` will try and load `server.js` (and fail), then fallback to `index.js`.
This is fine if you're using node, but I'm using [jspm](http://jspm.io/) which isn't too happy and errors with 'unable to find server.js'. Yes, this could be fixed on jspm's end, but it still leaves the unknown server.js reference in this package which might cause problems for other package managers (and even npm/node in the future?), so it'd be best to fix it here.


Schmavery:
Oh ok, looks like we forgot to remove that.  Thanks.

Issue #232, opened 2016-04-02
Sending giant emoji


jobbogamer:
Now that it's possible to change the thread emoji, is there some way that the giant one can be sent (the one you get when you press and hold the emoji)?

GAMELASTER:
Yea, it's possible. When I got some more time, I will implement to sendMessage a this feature!

Schmavery:
Going to close this because we can track it in #234

Issue #221, opened 2016-03-26
Sending audio files as an attachment


ICSandy:
I would like to send an audio file as an attachment. I attempted using api.sendMessage() to send the audio file attachment but I get an error. I can send attachments if it is a picture, but .wav or .mp3 files will not work and return this error:

```
ERR! Error in uploadAttachment TypeError: Cannot read property 'push' of undefined
ERR! Error in uploadAttachment     at /home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/src/sendMessage.js:208:50
ERR! Error in uploadAttachment     at Array.forEach (native)
ERR! Error in uploadAttachment     at /home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/src/sendMessage.js:202:15
ERR! Error in uploadAttachment     at /home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/src/sendMessage.js:46:9
ERR! Error in uploadAttachment     at tryCatcher (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:26:23)
ERR! Error in uploadAttachment     at Promise._settlePromiseFromHandler (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:507:31)
ERR! Error in uploadAttachment     at Promise._settlePromiseAt (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:581:18)
ERR! Error in uploadAttachment     at Promise._settlePromises (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:697:14)
ERR! Error in uploadAttachment     at Async._drainQueue (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:123:16)
ERR! Error in uploadAttachment     at Async._drainQueues (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:133:10)
ERR! Error in uploadAttachment     at Immediate.Async.drainQueues [as _onImmediate] (/home/pi/Desktop/botboy_test/test here/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:15:14)
[...]

PR #234, opened 2016-04-03
Adding a emoji support

GAMELASTER:
As a suggestion of #232, I did it. The sizes available are a `small` and `medium`, big isn't working.
It's available a one more size (a extra small), but I don't found a name of variable. (you can "spawn" only two sizes of emoji).

The "like" :+1: button isn't a emoji, it's a sticker, where
`369239263222822` is a small
`369239343222814` is a medium
`369239383222810` is a large


bsansouci:
Thanks for contributing. In the future could you not make a PR with two different changes? That would help us.

The sound part isn't quite right. Looking at #221 @ICSandy's change just uploads it as an audio file. We looked into it and committed this 9b2a5d3bfd05e5141165a9a565673ce7a23282ef.

For the emoji part, there are a ton of problems that need to be addressed before we can move forward. 
- some unicode emojis don't work on web and show this warning
screen shot 2016-04-03 at 20 25 38
- as you said, some emojis' large/medium look the same
- thumbs up is a sticker, not an emoji, so what happens when you try to give it a size? (Nothing happens)
- what happens when you try to send an emoji and an attachment? Should you be able to? (Answer: it works, and you should probably not be able

Also this function signature is really not great, as it closely follows what FB is doing. Can you change it to be `api.sendMessage({emoji: "🖕", emojiSize: "small"})`. The function should check that the keys `emoji` and `emojiSize` are _always_ provided together (and only provided together). It should also check that `emojiSize` is either `"small"`, `"medium"` or `"large"` and return an error if not. Lastly this should check that `emoji` is exactly one unicode character long (to prevent people from sending stickers instead).

GAMELASTER:
Sorry for inactive, im going to change it soon.

GAMELASTER:
@bsansouci 
So, I reverted my edits for not a merge problem and pushed it again.
I removed audio_ids patch.
To an emoji:
- The size of Like Sticker is about an Sticker ID, a big Like got another ID than a medium etc. So the sizing of the emoji don't do anything.
- With medium and large differences I think it's not a big problem
- Sending an attachment with emoji works nice, Example:

image
- Im little confused about throwing an errors in Handlers, I found one usage of error in handleUrl, but I dont think its a best way, maybe throwing an exception
- I have list of (almost) supported Facebook emojis, but I don't know if its a best idea to check if its valid emoji

src/sendMessage.js
```diff
@@ -182,6 +184,26 @@ module.exports = function(defaultFuncs, api, ctx) {
...
     cb();
   }
+  
+  function handleEmoji(msg, form, callback, cb) {
+    if (msg.emoji) {
+      if(msg.emojiSize == null)
+      {
+        return callback({error: "emojiSize property is empty"});
+      }
+      if(msg.emojiSize != "mini" && msg.emojiSize != "medium" && msg.emojiSize != "large")
```
Schmavery: mini?..

src/sendMessage.js
```diff
@@ -182,6 +184,26 @@ module.exports = function(defaultFuncs, api, ctx) {
     }
     cb();
   }
+  
+  function handleEmoji(msg, form, callback, cb) {
+    if (msg.emoji) {
+      if(msg.emojiSize == null)
+      {
+        return callback({error: "emojiSize property is empty"});
```
Schmavery: What about people who set emojiSize but not emoji?

src/sendMessage.js
```diff
@@ -182,6 +184,26 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+      }
+      if(msg.emojiSize != "mini" && msg.emojiSize != "medium" && msg.emojiSize != "large")
+      {
+        return callback({error: "emojiSize property is invalid"});
+      }
+      if(msg.emoji.length > 3)
```
Schmavery: Can we do a little better than this?

src/sendMessage.js
```diff
@@ -182,6 +184,26 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+      {
+        return callback({error: "Invalid emoji"});
+      }
+      form['message_batch[0][body]'] = msg.emoji;
```
Schmavery: Won't this silently overwrite the body of a message?  I'd like to give an error or at least a warning if someone tries to do this.

Schmavery:
Hey @GAMELASTER,
I'm not sure how you're getting this nice example of emoji+attachment.  It seems like when @bsansouci and I were trying it the other day, we had a lot of rendering issues in messenger.com and facebook.com
I agree that I wouldn't want to explicitly name every emoji that is supported by facebook in our code.
Just so it's clear, this explicitly doesn't support sending the standard thumbs up, is that correct?
Just some thoughts.

src/sendMessage.js
GAMELASTER: Sorry! I again pushed a wrong revision of code!

GAMELASTER:
Sorry for inactive! I will lookup this soon!

GAMELASTER:
Let's finish and close this one! Sorry for one year latency 🤣 

I tried to solve all the requests to change, there is some mine notices:
- Is hard to detect if the `emoji` really contains emoji
- If `emojiSize` is not set, it will send a classical size of emoji
- Is possible to send emoji (with all sizes) along with video and image (probably it will work on every other attachment)
- Another-skin emojis are working good too
- Default thumbs up emoji still works like sticker (every size is another sticker), but you can send a Thumbs Up emoji, but it looks like normal emoji

src/sendMessage.js
```diff
@@ -184,6 +186,26 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+      return callback({error: "emoji property is empty"});
+    }
+    if (msg.emoji) {
+      if (msg.emojiSize == null) {
+        msg.emojiSize = "medium";
+      }
+      if (msg.emojiSize != "small" && msg.emojiSize != "medium" && msg.emojiSize != "large") {
+        return callback({error: "emojiSize property is invalid"});
+      }
+      if ((form['body'] != null || form['body'] != "") == false) {
```
bsansouci: This if statement seems a bit weird. You want to error when the body _is_ empty right?
So maybe do:
if (form['body'] != null && form['body'] != "") 
Because emojis are only valid when the body is empty

bsansouci:
Seems good! I added just one comment.

src/sendMessage.js
GAMELASTER: Ah yes, you are right! Thanks and sorry, I did it at late-night, so my thinking has been little weird :-D I'm going change that.

GAMELASTER:
It's done! (forgot to write! :rofl: )

bsansouci:
Looks good! Thanks a lot 👍

PR #240, opened 2016-04-05
Fix issue where getThreadHistory returns extra message.

### LINKED ISSUE #239: getThreadHistory result length off by 1
### api.getThreadHistory(threadID, start, end, timestamp, [callback])

getThreadHistory always returns one extra message.

Example: `api.getThreadHistory("869042309831501", 0, 1, null, myFunc);` returns the 2 most recent messages when it should return 1.

The limit is calculated correctly as `end - start + 1` in `getThreadHistory.js` but Facebook returns limit+1 messages.

UsmannK:
Fixes #239


Schmavery:
Thanks!

PR #241, opened 2016-04-06
Update NPM package for getThreadInfo

nathanhleung:
I was wondering why getThreadInfo wasn't working and after checking out the repo it looks like it's because it's a recently added feature that hasn't been published to NPM yet. I've updated the version number to reflect the changes (minor update since features were added). Just run `npm publish` and you should be good!

(I'm not sure if you were purposely holding out on publishing or not - if so, I can always just find another way to get thread information)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "facebook-chat-api",
-  "version": "1.0.8",
+  "version": "1.1.0",
   "description": "Facebook chat API that doesn't rely on XMPP.  Will NOT be deprecated April 30th 2015.",
   "scripts": {
     "test": "node_modules/.bin/mocha",


Schmavery:
This feature is only in master currently... feel free to [sync to master](https://docs.npmjs.com/files/package.json#github-urls) if you want the bleeding edge.
The next update will have to be a major update because it contains breaking changes.

joepie91:
For future readers of this thread: Installing from Git is not a good idea - you can't use any semantic versioning ranges when installing from a commit ID (meaning you don't get updates), and don't have any API stability guarantees when installing from master.

It's best to just publish a patch/minor release to NPM as soon as possible, especially for breaking bugs or essential missing features.

Issue #15, opened 2015-06-09
Requires `request-promise` but does not depend on it.


jackwilsdon:
Line 4 of utils.js on version 2.0 still requires request-promise, however it is not included in the package.json. It looks like c54c677 was not included in the 2.0 `npm` release.

jackwilsdon:
So it turns out 67ef22 (version 2.0) was uploaded to npm before c54c677 was pushed to the repository (no longer requiring on request-promise), however it looks like request-promise was never a dependency of facebook-chat-api.

Schmavery:
Good catch, thanks.

PR #247, opened 2016-04-09
get/accept/ignored message requests

how2945ard:
based on https://github.com/Schmavery/facebook-chat-api/pull/193

add function to get current pending message requests, 

first params for acceptMessageRequest/ignoreMessageRequest are array of threadIDs


src/listen.js
```diff
@@ -160,6 +160,11 @@ module.exports = function(defaultFuncs, api, ctx) {
 
               return globalCallback(null, fmtMsg);
               break;
+            case 'inbox':
+              // Could be message request
```
Schmavery: Could be?  Can it also be other things?

src/listen.js
```diff
@@ -160,6 +160,11 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+              // Have to use getMessageRequests() for more infomation.
+              return globalCallback(null, v);
```
Schmavery: As you can see in the rest of this file, we try to avoid returning raw facebook objects whenever possible.  They have a habit of changing without warning and breaking the signature of the api.

src/getMessageRequests.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
```
Schmavery: It looks like this function is a copy/paste of getThreadList with `inbox` changed to `pending`.  Is there a reason we can't just add an optional boolean to getThreadList to select pending threads?

Schmavery:
Hey @how2945ard, thanks for your contribution!
I've left a couple of comments inline for you to consider.
As I asked in #193, is there a reason you didn't consider merging the accept and ignore functions into an `acceptMessageRequest(accept)` that takes true when you want to accept and false for ignore?
Can we get some basic docs for this?

how2945ard:
Sure!
Thanks for your comments and I will work on these later.

src/listen.js
how2945ard: It's kind of hard to reproduce the message request scenario.

v looks like this:
`
{ unread: 1,
  unseen: 1,
  seen_timestamp: 0,
  realtime_viewer_fbid: <fb_id>,
  type: 'inbox' }
`
I get these `v.type === 'inbox'` things whenever a message is received ( including message requests ).
I think there have to be some `v.type` that are for message requests only, but I haven't figure this thing out yet.

Schmavery: Understandable.  Maybe we don't need a listen event then for now?

src/getThreadList.js
```diff
@@ -4,23 +4,32 @@ var utils = require("../utils");
 var log = require("npmlog");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadList(start, end, callback) {
+  return function getThreadList(start, end, isPending, callback) {
```
Schmavery: Can we make the `pending` argument option so that we don't have to break backward compatibility?

src/getThreadList.js
```diff
@@ -4,23 +4,32 @@ var utils = require("../utils");
...
     if (end <= start) end = start + 20;
 
-    var form = {
-      'client' : 'mercury',
-      'inbox[offset]' : start,
-      'inbox[limit]' : end - start,
+    var messageBox = 'inbox';
+    if (isPending) {
+      messageBox = 'pending';
     };
```
Schmavery: I usually feel like 
`var messageBox = pending ? "pending" : "inbox";`
is more explicit.
I would have also probably called it `pending` rather than `isPending` (and accept rather than isAccept) but I guess that's getting reeeaally nitpicky.

src/handleMessageRequests.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+  return function handleMessageRequests(threadID, isAccept, callback) {
+    if(!callback && utils.getType(isAccept) === 'Function') {
```
Schmavery: I don't really understand why you're making this check in this way.  Why are you checking `!callback`?
Don't you just care about `utils.getType(isAccept) !== 'Boolean'`?
Nitpick: Can we make the name of this function singular?

src/handleMessageRequests.js
```diff
@@ -0,0 +1,48 @@
...
+    };
+
+    if (utils.getType(threadID) !== "Array") {
+      threadID = [threadID];
+    }
+
+    var messageBox = 'other';
+    if (isAccept) {
+      messageBox = 'inbox';
+    };
```
Schmavery: Same as the other comment, would be clearer as one line.

DOCS.md
```diff
@@ -359,14 +360,15 @@ __Arguments__
...
-### api.getThreadList(start, end, callback)
+### api.getThreadList(start, end, isPending, callback)
 
 Will return information about threads.
 
 __Arguments__
 
 * `start`: Start index in the list of recently used threads.
 * `end`: End index.
+* `isPending`: Boolean indicating that it will return the thread list of message requests.
```
Schmavery: Can you phrase this a little more clearly?
I think "Optional boolean, set to true if you want the list of message requests." conveys the behaviour more explicitly.
I'm also wondering if we should just call it like.. "requests" or something?  I'm not totally sure that pending is obviously associated with message requests.  Shouldn't matter too much with the explanation alongside though, so I don't really care.

src/getThreadList.js
how2945ard: Could you show me some tips on how to make `pending` optional and `callback` required at the same time? 
Maybe I can do some type check on `pending`?

how2945ard:
Need to make `pending` optional

src/getThreadList.js
```diff
@@ -4,22 +4,28 @@ var utils = require("../utils");
 var log = require("npmlog");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadList(start, end, callback) {
+  return function getThreadList(start, end, pending, callback) {
```
how2945ard: @Schmavery 
How should I make `pending` optional? Could you show me some tips

RobinHoutevelts:
Cool, this is very handy.

However, I'd also like to be able to listen for an event when someone initiates a chat request.
Right now, it seems ( I could be mistaken ) we have to poll `getThreadList()` with the `pending` boolean set to true in order to find out about new requests.

In the `parsePackets()` [function of **listen.js**](https://github.com/Schmavery/facebook-chat-api/blob/beae8ba16ad55f7bc1a763a4274989a8e6dee3f4/src/listen.js#L109) this object with type **inbox** get's passed.
```
{ 
  unread: 0,
  unseen: 0,
  seen_timestamp: 0,
  recent_unread: 0,
  realtime_viewer_fbid: 100012132393019,
  type: 'inbox' 
}
```

I've only been looking at this project for about an hour so I'm not sure if it's something useful.

src/getThreadList.js
RobinHoutevelts: You could insert this at line [#15](https://github.com/NTU-ArtFest22/facebook-chat-api/blob/795bdbc83323c98f546fa8084ed8b7eda10f100f/src/getThreadList.js#L15) to add backwards compatibility and setting `pending` to **false**.
```javascript
if(!callback && typeof pending === 'function') {
  callback = pending;
  pending = false; // Default pending to False
}
```

If you don't **everyone** has to update their scripts because of the extra parameter.

bsansouci: @RobinHoutevelts -> yup, that works (but use `utils.getType(pending) === 'Function'`) and don't do `!callback` but do `utils.getType(callback) === 'Undefined'`

AllanWang:
With reference to [QuickUnit's commits](https://github.com/QuickUnit/facebook-chat-api/commit/f6828953594ca6b26810e363161fe2b210943be3), if the part below was added instead, would that work?

```javascript
if(utils.getType(callback) === 'Undefined' && utils.getType(end) !== 'Number') {
      throw {error: "please pass a number as a second argument."};
}

if(utils.getType(callback) === 'Undefined') {
      if (utils.getType(pending) === 'Function') {
            callback = pending;
            pending = false; //default to inbox
      } else if (utils.getType(pending) !== 'Boolean') {
            throw {error: "please pass a Boolean as a third argument."};
      } else {
            throw {error: "getThreadList: need callback"};
      }
}
```

Schmavery:
This looks mostly right, though you could probably merge those two sections.
```js 
if(utils.getType(callback) === 'Undefined') {
      if (utils.getType(end) !== 'Number') {
          throw {error: "please pass a number as a second argument."};
      } else if (utils.getType(pending) === 'Function') {
            callback = pending;
            pending = false; //default to inbox
      } else if (utils.getType(pending) !== 'Boolean') {
            throw {error: "please pass a Boolean as a third argument."};
      } else {
            throw {error: "getThreadList: need callback"};
      }
}
```

Look good to you @bsansouci ?
If you wanted to resubmit this PR with the changes, feel free :smile:

AllanWang:
I will test it and submit a PR if it works for me.

AllanWang:
Also sidenote, but if utils.getType(callback) === 'Undefined' is better than !callback, I can change that for all the other js files

bsansouci:
Hey @AllanWang sorry we didn't answer for a while. If you want, you could fix the merge conflict and we could merge this in.

AllanWang:
@bsansouci if I'm not mistakened, we've [already added it](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#getThreadList) with a String parameter (request #301). I think this request can be closed.

bsansouci:
@AllanWang Good call. Closing now :)

PR #252, opened 2016-04-12
Update getThreadHistory.js

alexmylonas:
While experimenting with pages rather than users noticed minor typo on this file!


Schmavery:
Nice catch! Thanks.

PR #254, opened 2016-04-19
Update utils.js

AllanWang:
[] is invalid in openshift


AllanWang:
I'm not sure how you want to name the variable so I just did that. I have confirmed that this change allows openshift to load.

Schmavery:
This looks like more of a problem in openshift rather than our api.  Try to find a workaround to install a newer version of node or sync to a fork with your change :)

PR #281, opened 2016-05-27
Add fbash under 'Projects using this API'

avikj:
Fbash is a node module that is installed globally. The user can start it by running the command `fbash` and then enter their Facebook credentials to log in. After it has been started, the user can send messages to their own Facebook account to have them run as commands on the connected computer. Its on npm here: https://www.npmjs.com/package/fbash


Schmavery:
Thanks for the contribution!  I just glanced over your source code and it looks like you had been having issues with duplicate messages.  This was an issue we had a while ago and should hopefully be fixed now.

avikj:
@Schmavery Thanks for accepting my PR. Regarding the issue with duplicate messages, I just tried removing the filter for duplicate messages, and now every command is received twice. Was the issue that was fixed regarding messages sent to one's own Facebook account? (I think that it's caused because the api.listen callback is triggered once for a sent message, and again for when the message is received).

Schmavery:
hmm, maybe there's still an issue when selfListen is enabled.  Thanks for letting us know.

Issue #294, opened 2016-07-06
Mute/unmute chat


renxinhe:
I'm not sure how difficult/easy it is to toggle mute and unmute on a specific chat. I have a "mute timer" project in mind, and couldn't find an easy way to inject into the chat html/php on [facebook.com](url). I was so excited to find this API, but after a bit of looking around, it doesn't seem like there is a mute feature.

Schmavery:
It would be a matter of sniffing the packets that the browser sends to facebook.com when muting/unmuting the chat and then reproducing them.  I haven't looked into it, but I'd imagine it'd be doable.  I'd consider accepting a PR for a api.changeMutedStatus similar to api.changeArchivedStatus.

GAMELASTER:
Yeah, actually it's contains a Messenger.com
```
https://www.messenger.com/ajax/mercury/change_mute_thread.php?dpr=1
POST
thread_fbid:100000XXXXXX540
mute_settings:-1 //-1 nonstop, 0 = disable, 1800 = 30 minutes, 3600 = hour etc..
payload_source:mercury
```
Then just check some property in Listen event for filtering if it is muted

bsansouci:
@renxinhe you should definitely contribute and add this! This looks like a small addition. 
Here's what you'll want to do:
- copy `markAsRead.js` and make `muteChat.js`
- replace `https://www.facebook.com/ajax/mercury/change_read_status.php` with `https://www.messenger.com/ajax/mercury/change_mute_thread.php`
- make `form` contain the `thread_fbid` and the `mute_settings` which should be both arguments to the function (I'm guessing `mute_settings` is the number of seconds you want the chat to be muted).
- fix all discrepancies

I'm sure @GAMELASTER would love to help you, if this is your first contribution.

renxinhe:
@bsansouci thanks for the help!
This is my first time looking into open source. I've been trying to hard code some request over via Postman to mute a chat. No luck yet. But I'll look into `markAsRead.js` and see if I can pull something together.

PR #295, opened 2016-07-06
mute/unmute chat

renxinhe:
Closes Schmavery/facebook-chat-api#294.
Implemented mute/unmute API, `muteChat.js`, based on `markAsRead.js`.


src/muteChat.js
```diff
@@ -0,0 +1,33 @@
...
+module.exports = function(defaultFuncs, api, ctx) {
+  // muteSecond: -1=permanent mute, 0=unmute, 60=one minute, 3600=one hour, etc.
+  return function muteChat(threadID, muteSecond, callback) {
+    if(!callback) {
+      callback = function() {};
+    }
+
+    var form = {};
+    form["thread_fbid"] = threadID;
+    form["mute_settings"] = muteSecond;
```
bsansouci: Can you add those directly in the object
```js
var form = {
  thread_fbid: threadID,
  mute_settings: muteSeconds,
}
```

src/muteChat.js
```diff
@@ -0,0 +1,33 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+  // muteSecond: -1=permanent mute, 0=unmute, 60=one minute, 3600=one hour, etc.
+  return function muteChat(threadID, muteSecond, callback) {
```
bsansouci: Can you add an s to `muteSecond` so it's `muteSeconds`?

bsansouci:
This looks great! Minor fixes and then we can merge :D

bsansouci:
Oh two more things that I just realized. We call chats `thread`s. Can you update the name to be `muteThread`?
Also you need to add `"muteThread"` here https://github.com/Schmavery/facebook-chat-api/blob/fddb01ffaf0e9119f6b44f78c2ed38e892bdb324/index.js#L88

Schmavery:
Thanks for working on this.  If you could double check to make sure your example code does what it's meant to, that'd be awesome.

GAMELASTER:
Well, its okay! But is good to mention the status if the user is muted contains a `message` variable on `listen`. And I'm not sure what is the variable name + I think it is contains a time to when is it muted

renxinhe:
You welcome! I ran the example on a node terminal, and a friend got muted for a minute after sending me a message. 
Messenger.com correctly showed "Muted until 1:02 pm"; I received the message around 1:01 pm.

GAMELASTER:
@renxinhe I'm sorry, the information about mute contains threadlist_info.php request:
image

renxinhe:
Sorry @GAMELASTER , what's the issue here? I just looked at `api.listen`, and I noticed that certain message types like `file` and `photo` don't have a `threadID`. Is that what you were saying that I should handle?

GAMELASTER:
Ah, sorry! I just wanted to write into a documentation into a [getThreadList](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#getThreadList) how to get the mute state for thread, it is mute_settings variable on thread list info. But well, isn't in good format, maybe we can add some parsing for just return `mute_settings: true`.

renxinhe:
Oh I see. I can make an `api.getMuteStatus` function in another pull request if necessary.

Schmavery:
This would be added to the getThreadList response, not a new function, but sure, this can be added later.  Thanks for the investigation @GAMELASTER.

renxinhe:
Cool! Guess my job is done. It's curious to see that Facebook doesn't return a `threadID` for images and files. One'd assume everything is associated with a message thread.

bsansouci:
Awesome! Thanks a lot @renxinhe. Can you follow up with a PR to add `muteStatus` onto the thread object ([here](https://github.com/Schmavery/facebook-chat-api/blob/def4518900c2acf24d2e3ac0c84ce7ecd95fa859/utils.js)) whenever you get some time.

renxinhe:
Sure. Just add it to `function formatThread (data)` correct?

bsansouci:
Yes, sorry didn't copy paste the URL properly.

renxinhe:
@bsansouci The [thread object](https://github.com/Schmavery/facebook-chat-api/blob/master/utils.js#L621) already has a `muteSettings` attribute.

GAMELASTER:
@renxinhe well, yeah, it is, but it needs a parsing, look into mine image, its in format:
mute_settings:
  THREADID@facebook.com: CONSTANT
It will be nice to mute_settings just contains a value. I think enough will be just get a mute_settings first children variable value)

bsansouci:
@renxinhe What @GAMELASTER said ^. Also documentation is needed to explain `-1` and any other value :)

PR #300, opened 2016-07-18
Fixed read_receipt events

QuickUnit:
The 'read_receipt' and 'read' events were not properly handled. Now considering the 'messaging' type in the packet and handling them.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -28,6 +28,8 @@
 * [`api.sendTypingIndicator`](#sendTypingIndicator)
 * [`api.setOptions`](#setOptions)
 * [`api.setTitle`](#setTitle)
+* [`api.handleMessageRequest`](#handleMessageRequest)
+
 
 ---------------------------------------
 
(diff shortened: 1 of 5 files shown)


bsansouci:
Hey @QuickUnit thanks for contributing! But I'm pretty sure we're handling those right here: https://github.com/QuickUnit/facebook-chat-api/blob/c22a7200552c60af1bd2cf2ddf5d96f7a6ff99b7/src/listen.js#L171-L175. Maybe I'm misunderstanding?

QuickUnit:
Yeah, true indeed, I was working on the code I got from npm, and didn't check the repo. 

Perhaps release a new version?

bsansouci:
Wow I'm not sure how you got the npm version to work but you should never use it to make a PR. The npm version is the "production" version of the library, and in our case it's a copy of the github version but in lots of project those versions have build steps that do things (minification, strip annotations, dead code elimination etc...). 
The fact that our npm release is a bit behind shouldn't affect this ;) (though we'll try to get it done soon)

QuickUnit:
I just used it, and then fixed the code to make it work for me. After that I forked and applied the fix without reading the rest of the code, so I missed your change there. Nothing special here, just ignored some new additions. :)

PR #301, opened 2016-07-21
Add message request support

AllanWang:
Full credits to [QuickUnit's commit](https://github.com/QuickUnit/facebook-chat-api/commit/f6828953594ca6b26810e363161fe2b210943be3)

I changed the description a bit and changed the type checks as mentioned [here ](https://github.com/Schmavery/facebook-chat-api/pull/247#issuecomment-233505141).

I tested it and it seems to be working.

I did not modify the getThreadInfo file. The indents just changed.


src/getThreadInfo.js
```diff
@@ -4,49 +4,58 @@ var utils = require("../utils");
 var log = require("npmlog");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadInfo(threadID, callback) {
```
bsansouci: Can you re-indent this back to 2 spaces?

src/getThreadList.js
```diff
@@ -4,40 +4,54 @@ var utils = require("../utils");
 var log = require("npmlog");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadList(start, end, callback) {
```
bsansouci: re-indent to 2 spaces

src/getThreadList.js
```diff
@@ -4,40 +4,54 @@ var utils = require("../utils");
...
+            } else if (utils.getType(pending) !== 'Boolean') {
+                throw {
+                    error: "please pass a Boolean as a third argument."
+                };
+            } else {
+                throw {
+                    error: "getThreadList: need callback"
+                };
+            }
+        }
```
bsansouci: Here you can add 
```js
else if (utils.getType(pending) != "Boolean") {
  throw {
    error: "Please pass a boolean as a third argument."
  };
}
```

src/getThreadList.js
```diff
@@ -4,40 +4,54 @@ var utils = require("../utils");
...
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadList(start, end, callback) {
-    if(!callback && utils.getType(end) !== 'Number') {
-      throw {error: "please pass an number as a second argument."};
-    }
+    return function getThreadList(start, end, pending, callback) {
+        if (utils.getType(callback) === 'Undefined') {
+            if (utils.getType(end) !== 'Number') {
+                throw {
+                    error: "please pass a number as a second argument."
```
bsansouci: `Please` capital letter.

src/getThreadList.js
```diff
@@ -4,40 +4,54 @@ var utils = require("../utils");
...
+                };
+            } else if (utils.getType(pending) === 'Function') {
+                callback = pending;
+                pending = false; //default to inbox
+            } else if (utils.getType(pending) !== 'Boolean') {
+                throw {
+                    error: "please pass a Boolean as a third argument."
```
bsansouci: `Please` capital letter.

bsansouci:
Looks good. Ready to merge when you fix the nitpicking. Thanks a lot!

AllanWang:
@bsansouci 

```javascript
if (utils.getType(accept) !== 'Boolean') {
            throw {
                error: "Please pass a boolean as a second argument."
            };
        }
```

Should that be != instead of !== as well?

bsansouci:
Oh no it should always be `!==` and `===` _except_ when you're comparing with `null`. Then you'd want to use `!=` and `==`. So in this case use `!==`.

AllanWang:
Can you check now?

I also kept it as 

```javascript
      } else if (utils.getType(pending) != 'Boolean') {
        throw {
          error: "Please pass a Boolean as a third argument."
        };
      } else {
        throw {
          error: "getThreadList: need callback"
        };
      }
```
Since you'll need to throw the last callback error if there is no boolean and no callback

bsansouci:
Sounds good. Thanks.

Issue #283, opened 2016-06-07
List archived threads


matthiashhh:
As far as I can see, getThreadList only lists threads in the inbox. There are no way to list archived threads, I believe. I was able to get around this by changing 'inbox' to 'action:archived' in getThreadList .js code. It would be great if this functionality could be implemented.

Schmavery:
Please feel free to submit a PR where getThreadList accepts an optional boolean "archived" parameter.

AllanWang:
maybe we should change the pending boolean to a string parameter to support this. I will look into this later on.

Schmavery:
I think we have this now.  Thanks a lot @AllanWang !

PR #302, opened 2016-07-21
"type" for getThreadList

AllanWang:
changed pending boolean to String so that "archived" messages can be reached.

### DIFF
--- a/src/getThreadList.js
+++ b/src/getThreadList.js
@@ -24,7 +24,6 @@ module.exports = function(defaultFuncs, api, ctx) {
       }
     }
 
-    type = type.toLowerCase();
     if (type === 'archived') {
       type = 'action:archived';
     } else if (type !== 'inbox' && type !== 'pending') {
--- a/src/getThreadList.js
+++ b/src/getThreadList.js
@@ -13,7 +13,7 @@ module.exports = function(defaultFuncs, api, ctx) {
       } else if (utils.getType(type) === 'Function') {
         callback = type;
         type = 'inbox'; //default to inbox
-      } else if (utils.getType(type) != 'String') {
+      } else if (utils.getType(type) !== 'String') {
         throw {
           error: "Please pass a String as a third argument. Your options are: inbox, pending, and archived"
         };
--- a/src/getThreadList.js
+++ b/src/getThreadList.js
@@ -4,18 +4,18 @@ var utils = require("../utils");
 var log = require("npmlog");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function getThreadList(start, end, pending, callback) {
+  return function getThreadList(start, end, boxType, callback) {
     if (utils.getType(callback) === 'Undefined') {
       if (utils.getType(end) !== 'Number') {
         throw {
           error: "Please pass a number as a second argument."
         };
-      } else if (utils.getType(pending) === 'Function') {
-        callback = pending;
-        pending = false; //default to inbox
-      } else if (utils.getType(pending) != 'Boolean') {
+      } else if (utils.getType(boxType) === 'Function') {
+        callback = boxType;
+        boxType = 'inbox'; //default to inbox
+      } else if (utils.getType(boxType) != 'String') {
         throw {
-          error: "Please pass a Boolean as a third argument."
+          error: "Please pass a String as a third argument. Your options are: inbox, pending, and archived"
         };
       } else {
         throw {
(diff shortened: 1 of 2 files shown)


src/getThreadList.js
```diff
@@ -24,16 +24,23 @@ module.exports = function(defaultFuncs, api, ctx) {
       }
     }
 
-    if (end <= start) end = start + 20;
+    boxType = boxType.toLowerCase();
+    if (boxType == 'archived') {
```
bsansouci: use `===` and below `!==`

bsansouci:
I'm trusting that you tested `action:archived` :)

AllanWang:
@bsansouci yes I have tested this, and this was taken from [this suggestion.](https://github.com/Schmavery/facebook-chat-api/issues/283).

And just wondering, but how do you guys manage to find the new options such as action:archived? As far as I know there isn't a graphAPI for this that lists all the available branches.

Schmavery:
If you make the request in a browser while listening to the network traffic, you can see which GET and POST requests are being made to the facebook server by the facebook browser client.  Through a combination of this and reverse engineering the facebook browser client code, it's generally possible to figure out how most things work.

AllanWang:
@Schmavery do you have an example of a request you made to find a previous feature (say chat colors) and how you listened to it? If I learn how to do this I might be able to help with future features.

Schmavery:
Basically the idea is to open up facebook.com or messenger.com in your browser (let's use chrome as an example).  Then open up the  [dev tools](https://developer.chrome.com/devtools) and select the network tab.  It will probably be helpful for you to select filtering by XHR as indicated in this screenshot:
image
You then do whatever action you want to copy in the browser and see what requests come up in the network monitor.  You can click these and see what data is being sent in the request and which url it is targeting, what the response is, etc.  Then we simply try to emulate what is happening there in the API.

For changing thread color, you should see something like this when making the request in messenger.com:
image
If you compare this to [changeThreadColor()](https://github.com/Schmavery/facebook-chat-api/blob/master/src/changeThreadColor.js), you'll see that we're making a request with 'thread_or_other_fbid' and 'color_choice' just like the browser request is doing.  Some of the other fields like _user, _req, _a, fb_dtsg, tt_stamp etc. are added by our post function automatically because they are required by all requests that facebook makes.

AllanWang:
Thanks. Once you guys accept (or reject) this pull request I will look into sending emojis.

Schmavery:
In principle I have no problem with adding this functionality except that it's a breaking change and relies on entering keyword strings.  The former should be able to be fixed by checking if the boxType field is a boolean and reverting the the old behaviour in this case.  As a sidenote, I think we can simply call this argument 'type' rather than box type.

I'm never a fan of entering strings to get functionality, but in JS we unfortunately don't have many options there.  I like to require very exact input and error on everything invalid in these situations so that we don't end up with weird degenerate cases.  I'm glad you're checking whether boxType is one of the accepted values for this reason (though you could probably move that check up to [here](https://github.com/Schmavery/facebook-chat-api/pull/302/files#diff-017a6e261d15146e15fe740821e4aa47R16)). We should also remove the toLowerCase() [here](https://github.com/AllanWang/facebook-chat-api/blob/7359d6e26afe647409b8746201bb0df123d980b4/src/getThreadList.js#L27).  

Other than that, apart from a "!=" instead of a "!==" [here](https://github.com/Schmavery/facebook-chat-api/pull/302/files#diff-017a6e261d15146e15fe740821e4aa47R16), it looks ok to me.  @bsansouci ?

bsansouci:
Sounds good to me.

AllanWang:
@Schmavery since that line might be null, he said to use !=. That's why it will throw the no callback error. In regards to the boolean, I was hoping to push this out quickly so no one updated to the change. It was updated less than 24 hours ago anyways.

Schmavery:
Re: boolean: totally fair, I blanked for a second and didn't realize that was only just checked in.

Schmavery:
Maybe @bsansouci can clarify what he meant about !=, I think there's some misunderstanding there.  As long as you remove the toLowerCase, I'm good with this PR.

bsansouci:
@AllanWang Oh I think I didn't explain this very well. If you're going to compare _against null_, then you should use `==` and `!=` so something like `someVariable == null` is better than `someVariable === null` because in the latter case, if `someVariable` is `undefined` the triple equal will return false but the double equal will properly type cast and return true. If you're comparing with anything else than null, then use `===` and `!==`.

AllanWang:
@bsansouci what was the reasoning behind `utils.getType(type) != 'Boolean'` then? I'm assuming the getType method will always return something, be it boolean, string, or undefined. Either way, if you feel !== is more appropriate for that line I can change it. I'm still relatively new to Javascript and never really used === in any situation.

Schmavery:
Fair enough to think to copy what was there, but that should have been !== too haha.
The getType method should always be a string.

9.5 times out of 10, you're going to want to use === and !== (at least the way I like to code).  If you use == or !=, javascript will attempt to perform a type coercion as part of the equality check.  That is, if I'm checking `"" == false`, that will return true because `""` becomes `false`.  As you can imagine, you can get some strange behaviour in a hurry doing this.  For that reason, I prefer to be precise with my equality checks.  Checks using === and !== will also make sure the types of the things you're comparing match each other.

Checking against null is a bit of a special case, where js has `undefined` and `null` which are similar in some respects but not the same type.  That is, `null === undefined` is false.
Ben's suggestion is that if you want to check if something is null or undefined, you can use `foo == null`.  This is equivalent to checking `foo === null || foo === undefined`.

Hope that helps clear some stuff up!

AllanWang:
Changed accordingly

PR #307, opened 2016-07-28
Added customTagNames to formatThread

matthiashhh:
I need to retrieve the tags of the threads, thus added this to formatThread() in utils.js.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -370,7 +370,7 @@ __Arguments__
 * `start`: Start index in the list of recently used threads.
 * `end`: End index.
 * `type`: Optional String, can be 'inbox', 'pending', or 'archived'. Inbox is default.
-* `callback(err, arr)`: A callback called when the query is done (either with an error or with an confirmation object). `arr` is an array of thread object containing the following properties: `threadID`, <del>`participants`</del>, `participantIDs`, `formerParticipants`, `name`, `snippet`, `snippetHasAttachment`, `snippetAttachments`, `snippetSender`, `unreadCount`, `messageCount`, `imageSrc`, `timestamp`, `serverTimestamp`, `muteSettings`, `isCanonicalUser`, `isCanonical`, `canonicalFbid`, `isSubscribed`, `rootMessageThreadingID`, `folder`, `isArchived`, `recipientsLoadable`, `hasEmailParticipant`, `readOnly`, `canReply`, `composerEnabled`, `blockedParticipants`, `lastMessageID`.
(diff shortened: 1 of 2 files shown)


Schmavery:
Ok we can probably do this, but can you please give more information about what this is?  I'm not aware of tag names for threads.. maybe a feature I didn't hear about?  What format is the data in? If you're going to add this, can you also tweak the docs to reflect it?

matthiashhh:
The tags are called labels in the GUI and documentation:
https://www.facebook.com/help/935707876542654
In a thread object, the following member appears (example):
```
"page_thread_info": {
	"flagged": false,
	"customTagNames": ["My label"]
}
```
[Full example](https://github.com/Schmavery/facebook-chat-api/files/389189/example.json.txt)

matthiashhh:
Added  'customTagNames' to the list of properties in the callback from getThreadList in DOCS.md

Schmavery:
Does this only happen messages sent by pages?  What happens if you're just sending/receiving as a user account?  I would imagine the `page_thread_info` field wouldn't exist in this case, causing `formatThread()` to throw an exception?

matthiashhh:
You are right, the `page_thread_info` field is not there for messages in a user account. Moreover, when there are no `customTagNames`, the value is an empty array:
`"page_thread_info": []`
Is there a way around this? For those who use Facebook Pages, labeling is a very useful feature.

Schmavery:
I think I would be okay with simply only including this field for pages.  If it is a normal user message (i.e. no `page_thread_info` field, simply do not include that field in the returned object? 
@bsansouci let me know what you think of this.

Regarding the second issue, it looks like `page_thread_info` is sometimes an object and sometimes an empty array?  What does the `flagged: false` refer to in the original example?  Could this be something we could either strip out or default to false if unspecified?  Maybe I'm misunderstanding what's going on.

matthiashhh:
Here is a screenshot showing how to flag a thread:
flag
 I assume the `flagged` field refers this. Also you can see the labels (tags) 'Joshua' and 'Voice'.

matthiashhh:
The retrieval of customTagNames now handles undefined page_thread_info objects. In these cases, `customTagNames: undefined` will be returned. The same happens when no labels has been assigned a page thread. Have tested it against user threads, as well as page threads with and without labels.

You might want to take a look at my way of checking for undefined, I'm not a javascript programmer. It works, but for all I know, there could be better ways :)

matthiashhh:
Also, I guess we should give some thought to whether it is best to flatten the structure by adding the page_thread_info children directly to the root, as done in this request, or whether we should maintain the original structure by letting customTagNames (and flagged) be children of a pageThreadInfo object.

bsansouci:
Hey @matthiashhh, can you explain the difference between this PR and the one I just merged https://github.com/Schmavery/facebook-chat-api/pull/308 ?

Schmavery:
I'm going to close this, as it seems like #308 fixed the same issue.  Thanks for much for investigating this @matthiashhh, and please comment if there's anything we missed :)

PR #308, opened 2016-08-02
Adding tags to formatMessage()

matthiashhh:
The message objects have a member called 'tags'. The value of this member is an array.
Example:
`"tags": ["pages_comm_instant_reply","cg-enabled","sent","source:titan:web","inbox"]`
Example use case: For me, the "pages_comm_instant_reply" tag is important to identify, as this shows that the message is an Instant Reply:
https://www.facebook.com/help/1615627532020480

Tested on user threads and page threads.


bsansouci:
Hey @matthiashhh Sorry we didn't comment on this for... 6months :|
This looks good. Merging :)

Schmavery:
@bsansouci we need to remember to add docs

PR #315, opened 2016-08-17
sendMessage fix regarding fb api change

cookkkie:
https://github.com/Schmavery/facebook-chat-api/issues/311

Issue #311, opened 2016-08-12
Facebook changed sendMessage endpoint


Theblood:
I have this problem 
@Schmavery  :'(
Code:
```
book-chat-api/node_modules/bluebird/js/main/method.js:31:24)
ERR! ERROR in sendMessage -->      at /root/olaf-bot/node_modules/facebook-chat-api/utils.js:518:24
ERR! ERROR in sendMessage -->      at tryCatcher (/root/olaf-bot/node_modules/facebook-chat-api/node_modules
/bluebird/js/main/util.js:26:23)
ERR! ERROR in sendMessage -->      at Promise._settlePromiseFromHandler (/root/olaf-bot/node_modules/faceboo
k-chat-api/node_modules/bluebird/js/main/promise.js:507:31)
ERR! ERROR in sendMessage -->      at Promise._settlePromiseAt (/root/olaf-bot/node_modules/facebook-chat-ap
i/node_modules/bluebird/js/main/promise.js:581:18)
ERR! ERROR in sendMessage -->      at Promise._settlePromises (/root/olaf-bot/node_modules/facebook-chat-api
/node_modules/bluebird/js/main/promise.js:697:14)
ERR! ERROR in sendMessage -->      at Async._drainQueue (/root/olaf-bot/node_modules/facebook-chat-api/node_
modules/bluebird/js/main/async.js:123:16)
ERR! ERROR in sendMessage -->      at Async._drainQueues (/root/olaf-bot/node_modules/facebook-chat-api/node
_modules/bluebird/js/main/async.js:133:10)
ERR! ERROR in sendMessage -->      at Immediate.Async.drainQueues [as _onImmediate] (/root/olaf-bot/node_mod
ules/facebook-chat-api/node_modules/bluebird/js/main/async.js:15:14)
ERR! ERROR in sendMessage -->      at tryOnImmediate (timers.js:534:15)
ERR! ERROR in sendMessage -->      at processImmediate [as _immediateCallback] (timers.js:514:5)
ERR! ERROR in sendMessage -->   [Error: parseAndCheckLogin got status code: 404. Bailing out of trying to pa
rse response.]
info Got answer in  395
info Got answer in  2693
info Got answer in  4620
info Got answer in  4260
info Got answer in  4830
info Got answer in  297
info Got answer in  1707
info Got answer in  6831
```
[...]

Issue #278, opened 2016-05-18
GetOnlineUsers broken


tomexsans:
```
info Logging in...
info Logged in
info Request to reconnect
info Request to pull 1
info Request to pull 2
info Request to thread_sync
info Done logging in.
ERR! Error in getOnlineUsers Error: parseAndCheckLogin got status code: 404. Bailing out of trying to parse response.

ERR! Error in getOnlineUsers     at C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\utils.js:538:42

ERR! Error in getOnlineUsers     at tryCatcher (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\util.js:26:23)

ERR! Error in getOnlineUsers     at Function.Promise.attempt.Promise.try (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\method.js:31:24)

ERR! Error in getOnlineUsers     at C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\utils.js:523:24

ERR! Error in getOnlineUsers     at tryCatcher (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\util.js:26:23)

ERR! Error in getOnlineUsers     at Promise._settlePromiseFromHandler (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\promise.js:507:31)

ERR! Error in getOnlineUsers     at Promise._settlePromiseAt (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\promise.js:581:18)

ERR! Error in getOnlineUsers     at Promise._settlePromises (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\promise.js:697:14)
ERR! Error in getOnlineUsers     at Async._drainQueue (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\async.js:123:16)

ERR! Error in getOnlineUsers     at Async._drainQueues (C:\Users\tomex\Desktop\check\node_modules\facebook-chat-api\node_modules\bluebird\js\main\async.js:133:10)

[...]

PR #316, opened 2016-08-17
update to improve compatibility with new api

Mini-Geek:
Related to bug #311. Pull request #315 only attempted to fix sendMessage, this includes 3 more services.

Functions updated:

- addUserToGroup
- changeGroupImage
- sendMessage
- setTitle

Changes made:

- Remove `message_batch[0]`
- Change https://www.facebook.com/ajax/mercury/send_messages.php to https://www.facebook.com/messaging/send/

The unit testing currently produces 30 passes and 5 failures for me.

- Login As Page:
  1. should send text message object (user)
  2. should send sticker message object (user)
  3. should send basic string (user)
  4. should get a list of online users
- Login:
  5) should get a list of online users

The errors indicate:
- The URL for getOnlineUsers, https://www.facebook.com/ajax/chat/buddy_list.php, also returns a 404. I haven't been able to find the new URL and/or formatting for this.
- Things in Login As Page that send messages are not sent with an error saying "This person isn't receiving messages from you right now." (this may just be an issue with my test users/pages)

### DIFF
--- a/src/addUserToGroup.js
+++ b/src/addUserToGroup.js
@@ -24,29 +24,29 @@ module.exports = function(defaultFuncs, api, ctx) {
     var messageAndOTID = utils.generateOfflineThreadingID();
     var form = {
       'client' : 'mercury',
-      'message_batch[0][action_type]' : 'ma-type:log-message',
-      'message_batch[0][author]' : 'fbid:' + ctx.userID,
-      'message_batch[0][thread_id]' : '',
-      'message_batch[0][timestamp]' : Date.now(),
-      'message_batch[0][timestamp_absolute]' : 'Today',
-      'message_batch[0][timestamp_relative]' : utils.generateTimestampRelative(),
-      'message_batch[0][timestamp_time_passed]' : '0',
-      'message_batch[0][is_unread]' : false,
-      'message_batch[0][is_cleared]' : false,
-      'message_batch[0][is_forward]' : false,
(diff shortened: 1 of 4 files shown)


Kurimizumi:
`changeThreadEmoji()` is also broken, but `changeThreadColour()` is still working.

If I find any other broken functions I'll post them too.

Schmavery:
Thanks, this is awesome.
In the past, sometimes pages use an older version of the service, it seems.  Is there any chance the pages tests pass with no changes?
GetOnlineUsers is a known issue and the api function probably has to be deprecated https://github.com/Schmavery/facebook-chat-api/issues/278

Mini-Geek:
Kurimizumi, changeThreadEmoji is working for me. I'm not sure what problems you're having with it. 

I just checked, my changes didn't break the Login As Page tests. They didn't fix any of them, either: the same 4 tests fail with or without my changes here (again, that's with my test setup, which may be the issue).

GAMELASTER:
I think we need update a readme too for inform others which version to use if some function dont work

Schmavery:
Thanks again @Mini-Geek, I'll push an update to npm when I get home today.

Mini-Geek:
Awesome, glad I could help! Also, I figured out that for a page to message someone, they have to message the page first. So I messaged my page as my test user, and disabled the two "should get a list of online users" tests, and now I get no failures when I run the test.

So it all appears good to me! (or, as good as it can with that known broken API)

PR #318, opened 2016-08-18
remove unit test references to getOnlineUsers

Mini-Geek:
As of 1.2.0, the unit test still references the getOnlineUsers api removed due to #278. These tests should be removed.


Schmavery:
Haha yeah, realized last night that I forgot that one.  Thanks.

PR #320, opened 2016-08-19
update dependencies

ZakariaRidouh:


Schmavery:
Is it complaining when installing or something?
Can you delete node_modules after making this change, npm install, then run the tests to make sure this doesn't mysteriously break anything?
Thanks! :smile:

bsansouci:
Hey I don't think we need to update this unless you want to use some new functionality. I'll close this PR but feel free to reopen it.

PR #325, opened 2016-09-07
Fix: Avoid bad rendering of README on npm

cadesalaberry:


Schmavery:
Will this fix the problem?  How is npm getting confused?

PR #327, opened 2016-09-08
Fix: Avoid bad rendering of README on npm

cadesalaberry:
Sorry, the last commit did not go through completely.

I tested it with `marky-markdown`, the problem was that the `a` tag  was not closed.

I closed it and moved it after the title.

For clarification, I was talking about the html rendering of the `README.md` file on [npm](https://github.com/Schmavery/facebook-chat-api)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This API is still the only way to automate chat functionalities on a user accoun
 
 _Disclaimer_: We are not responsible if your account gets banned for spammy activities such as sending lots of messages to people you don't know, sending messages very quickly, sending spammy looking URLs, logging in and out very quickly... Be responsible Facebook citizens.
 
-See [below](#projects) for projects using this API.
+See [below](#projects-using-this-api) for projects using this API.
 
 ## Install
 ```bash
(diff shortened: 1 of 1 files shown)


Schmavery:
Huh well.. the `a` tag was closed... I guess npm's parser sucks.

bsansouci:
O_O
Can you file an issue on https://github.com/npm/marky-markdown to tell them they probably should support more than a subset of HTML

Schmavery:
http://stackoverflow.com/questions/3558119/are-self-closing-tags-valid-in-html5
I guess you can't self-close an anchor tag... pretty lame.

Schmavery:
It looks like the header auto-creates an anchor tag.  Can we link to #projects-using-this-api instead and remove the `<a name="projects" />` instead?

cadesalaberry:
@bsansouci  Yep, that's what I realised about anchor tags too...

@Schmavery That's what I wanted to do, but I realised you could have used the link somewhere else apart from github.

I figured it would be better to keep it that way.

However if the `README.md` is the sole place the **#projects** is used, yes it can be referenced with **#projects-using-this-api**.

Schmavery:
Let's go with #projects-using-this-api. I'm pretty sure nothing else references it, and if so, it will still point to the same page.

cadesalaberry:
Roger roger.

And done.

PR #330, opened 2016-09-11
Update formatPresence

ALiangLiang:
The Presence.statuses is broken for a long time. I change this to "online" to check is this user online. It's a boolean value represents is this user online. Or will be undefined when occur a unknown situation for API.


Schmavery:
What other values are possible for presence.ol?  Have you tried logging in from a phone etc?  I'd also prefer not to have a key that points to undefined, I'd rather have a default value or throw an exception if really necessary.

ALiangLiang:
#362 solved the problem too.

Schmavery:
thanks for letting us know #362 worked for you too.  I'll merge that I guess, thanks for following up.

Issue #329, opened 2016-09-10
api.markAsRead fails silently


Mini-Geek:
Maybe the Facebook API changed. The unit test still passes, but I don't see it reflected by the Facebook site or Android Messenger app. The request sent by the current site includes the following params, and this API only includes `ids[threadId]`.

ids[REMOVED]=true
watermarkTimestamp=1473520952998
titanOriginatedThreadId=REMOVED
shouldSendReadReceipt=true
commerce_last_message_type=non_ad

Schmavery:
Interesting, thanks.  Can you try adding these params to our call and seeing if that fixes it?

Mini-Geek:
I tried that, the one parameter I'm struggling to understand is titanOriginatedThreadId. Its format appears to match the unused utils.generateThreadingID method. Can you tell me about that method? What is the clientID that should be passed in? Here's what I've got so far:

    form["ids[" + threadID + "]"] = true;
    form["watermarkTimestamp"] = new Date().getTime();
    form["shouldSendReadReceipt"] = true;
    form["commerce_last_message_type"] = "non_ad";
    form["titanOriginatedThreadId"] = utils.generateThreadingID("abcd");

This causes a loop of "WARN parseAndCheckLogin: Got status code 500 retrying...". Looking into it, it's a generic HTML error page with a "Sorry, something went wrong." message, so I'm guessing Facebook doesn't like my clientID value, or something else about the request.

Schmavery:
Can you try ctx.clientID

Mini-Geek:
Adding ctx.clientID lets the call go through. But with or without the four new params, the read notifications are working in a few chats (both individual and group) I have going, but not working in one particular group chat. However, I noticed that read notifications are not working through the official app in this particular chat, either. I'm satisfied that this is currently a bug with Facebook and not with your API.

I still don't know what the issue is with this group chat and read notifications, but I think this issue can be closed for now.

Schmavery:
Thanks for investigating this, @Mini-Geek, it really helps a lot.
Hopefully facebook can get this cleared up on our end and it's good to have some initial investigation into a fix.
Does the group chat for which it fails succeed when you don't add the parameters?  If the parameters make markAsRead strictly better, I'd love to get a pull request so that we can add them to our api.

Mini-Geek:
[...]

PR #332, opened 2016-09-13
markAsRead has additional parameters

Mini-Geek:
Pull request from issue #329. All unit tests pass.

Note that I have not noticed that these changes are necessary for markAsRead to function as intended. The issue with #329 only happens on one thread that I know of, and appears to be an issue with Facebook, not this API. While investigating, I found that the official site included more parameters than this API did. These parameters may become important at some time in the future.


Schmavery:
When possible, I think it's useful to be closer the the message that they actually send :)  Thanks!

PR #355, opened 2016-11-13
allow node versions > 4

ZakariaRidouh:
this fixes compatibility with yarn which doesn't allow you to install packages that don't allow your current node version

cheers! :smile:


Schmavery:
Thanks @ZakariaRidouh!  I can't see any problem with this and always happy to better support yarn :)

PR #360, opened 2016-12-15
getThreadHistory: Filter out messages containing error attachments

### LINKED ISSUE #357: Attachments sometimes returned as 'error'
Looks like sometimes when fetching message history, the facebook api will return an error for one of the attachments in a message, which throws an error in the api here: 
https://github.com/Schmavery/facebook-chat-api/blob/master/utils.js#L338 

This is the contents of the stringified attachment: `{"attach_type":"error"}`

I am consistently returned about 17 of these while fetching ~72,000 messages, although a week ago fetching the same messages didn't return any errored attachments.

IMO it would be better for the API to gracefully continue without throwing an error, potentially removing the attachment from the message (or keep it).

ivkos:
Fixes #357

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -228,6 +228,16 @@ function getGUID() {
   return id;
 }
 
+function filterOutErrors(message) {
+  var hasErrors = message.attachments.some(att => att.type === "error");
+
+  if (hasErrors) {
+    log.warn("Message contains error attachments", message);
+  }
+
+  return !hasErrors;
+}
+
 function _formatAttachment(attachment1, attachment2) {
   // TODO: THIS IS REALLY BAD
   // This is an attempt at fixing Facebook's inconsistencies. Sometimes they give us
@@ -333,6 +343,15 @@ function _formatAttachment(attachment1, attachment2) {
         width: attachment1.metadata.dimensions.width,
         height: attachment1.metadata.dimensions.height,
         duration: attachment1.metadata.duration,
+      };
+    case "error":
+      return {
+        type: "error",
+
+        // Save error attachments because we're unsure of their format,
+        // and whether there are cases they contain something useful for debugging.
+        attachment1: attachment1,
+        attachment2: attachment2
       };
     default:
       throw new Error("unrecognized attach_file `" + JSON.stringify(attachment1) + "`");
@@ -677,6 +696,7 @@ module.exports = {
   parseAndCheckLogin: parseAndCheckLogin,
   saveCookies: saveCookies,
   getType: getType,
+  filterOutErrors: filterOutErrors,
   formatMessage: formatMessage,
   formatDeltaMessage: formatDeltaMessage,
   formatEvent: formatEvent,
--- a/utils.js
+++ b/utils.js
@@ -228,16 +228,6 @@ function getGUID() {
   return id;
 }
 
-function filterOutErrors(message) {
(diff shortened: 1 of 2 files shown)


Schmavery:
Thanks for taking a look at this issue.  I'm not totally sure how this solves any problems, were you able to replicate the problem in #357?  I had thought the exception was being thrown from formatMessage.

In an ideal world would it be nicer if we could just format these errors like we do in listen?

ivkos:
Yes, I do get the same messages with error attachments.

The exception is thrown from _formatAttachment because it doesn't handle
attach_type=error. The PR fixes that by handling this type of attachment
and formatting it, while saving the original attachment(s). As far as I've
tested, the original attachment doesn't contain any useful information
about the error, but I thought it might be a good idea to save them just in
case this changes or behaves differently on other accounts.

The PR also filters out messages that contain error attachments, so they
don't get returned in the result of getThreadHistory. If you want me, I can
leave getThreadHistory as it is, so that it returns everything including
messages with error attachments.



On Dec 16, 2016 04:12, "Avery Morin" <notifications@github.com> wrote:

Thanks for taking a look at this issue. I'm not totally sure how this
solves any problems, were you able to replicate the problem in #357
<https://github.com/Schmavery/facebook-chat-api/issues/357>? I had thought
the exception was being thrown from formatMessage.

In an ideal world would it be nicer if we could just format these errors
like we do in listen?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Schmavery/facebook-chat-api/pull/360#issuecomment-267499909>,
or mute the thread

Schmavery:
Oh I see now.  Thanks for clarifying!
@bsansouci what do you think, are we good to merge?

AlexGustafsson:
Are there a way of suppressing the warnings thrown?

ivkos:
@AlexGustafsson Yeah, you can set the log level to something higher than "warn", i.e. "error" or "silent". 
```js
api.setOptions({ logLevel: "error" });
```
See [api.setOptions](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#setOptions) for more info.

@Schmavery @bsansouci Do you think the "warn" level is appropriate for these messages? Should we use something lower, e.g. "info"?

Schmavery:
Warn seems fine to me.
ping @bsansouci merge?

utils.js
```diff
@@ -334,6 +344,15 @@ function _formatAttachment(attachment1, attachment2) {
         height: attachment1.metadata.dimensions.height,
         duration: attachment1.metadata.duration,
       };
+    case "error":
+      return {
+        type: "error",
+
+        // Save error attachments because we're unsure of their format,
+        // and whether there are cases they contain something useful for debugging.
```
bsansouci: Do errors really not have a specific format? Did you encounter those in the wild? If you would you mind sharing the shape of the object.

bsansouci:
@ivkos Hey, thanks for the PR.
Am I understanding correctly that you're _removing_ messages from the return of `getThreadHistory` if they had an attachment that has an error? That doesn't seem like a good behavior, as it seems like there would be no way for the user of this library to retrieve those specific message. I don't think we should filter the messages in this case.

Supporting the `error` type seems like a good thing though!

Schmavery:
Thanks!

Issue #362, opened 2016-12-28
updatePresence returning 'undefined' for 'statuses' field


lyneca:
With the `updatePresence` option on, `api.listen()` returns the following when a user comes online:

```
{ type: 'presence',
  timestamp: [timestamp],
  userID: '[user ID]',
  statuses: undefined }
```

when [in the docs](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#presence) it says that it should return
>An object with the following fields: fbAppStatus, messengerStatus, otherStatus, status and webStatus. All can contain any of the following values: "active", "idle", "invisible", "offline".

Is this a bug, am I doing something wrong, or is it just FBM being frivolous?

lyneca:
I think the error is at [line 651](https://github.com/Schmavery/facebook-chat-api/blob/master/utils.js#L651) of `utils.js`.

```js
function formatPresence(presence, userID) {
  return {
    type: "presence",
    timestamp: presence.la * 1000,
    userID: userID,
    statuses: presence.a // <--
  };
}
```
I did some snooping in the messenger.com and facebook.com requests and there isn't a `p` key in the presence response for the numbers - only a numeric `a` key which I believe is 0 for active and 2 for idle/away. I haven't tested it yet though, mostly because I don't know how to test local changes to node.js packages 😛 

Here's a response I got when someone became online:
`for (;;); {"t":"msg","seq":3139,"u":[user id],"ms":[{"overlay":{"[user id]":{"la":1482908493,"vc":10,"s":"push","a":2}},"type":"buddylist_overlay"}]}`

And here's one I got when someone went idle (after five minutes):
`for (;;); {"t":"msg","seq":3160,"u":[user id],"ms":[{"overlay":{"[user id]":{"la":1482908493,"vc":10,"s":"push","a":0}},"type":"buddylist_overlay"}]}`

lyneca:
It looks like facebook no longer sends seperate data for 'active on facebook.com', 'active on messenger.com', 'active on mobile' etc and just combines them into a single 'is user active parameter'.

lyneca:
yep that fixed it, I'll send a pull request with an update for the docs.

ALiangLiang:
Excuse me. What is 'presence.a' means? It's ever seen 0 and 2. :)

PR #363, opened 2016-12-28
Fix for #362

lyneca:
Updated to follow a change in the way Facebook shows user presence on Messenger

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -526,8 +526,6 @@ If enabled through [setOptions](#setOptions), this will also return presence, (`
 - `type`: The string "presence".
 - `timestamp`: How old the information is.
 - `userID`: The ID of the user whose status this packet is describing
-<!-- The form of the status request response has changed; it no longer sends five seperate keys - instead it sends a single value. The status states have been condensed to three numbers. -->
-<!-- - `statuses`: An object with the following fields: `fbAppStatus`, `messengerStatus`, `otherStatus`, `status` and `webStatus`. All can contain any of the following values: `"active"`, `"idle"`, `"invisible"`, `"offline"`. -->
 - `statuses`: The online status of the user. `0` means the user is idle (away for 2 minutes) and `2` means the user is online.
 
 __Example__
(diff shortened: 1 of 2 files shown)


Schmavery:
Thanks for investigating this.  You can delete those lines in the docs rather than commenting them out. 
Are there any other possible values for presence?

lyneca:
Well I would expect that there's a `1` state, possibly meaning invisible or something, but I didn't observe it while I was testing. I don't know how to activate an invisible state anyway so I can't test if that's what it is.

Schmavery:
This seems to be a different fix for the same issue as #330.  Care to comment as to why you think using presence.a is a better option?  Did the other fix not work for you?  Just curious and trying to figure out which one to merge.  Thanks and sorry for the delay in response!

DOCS.md
```diff
@@ -526,7 +526,9 @@ If enabled through [setOptions](#setOptions), this will also return presence, (`
 - `type`: The string "presence".
 - `timestamp`: How old the information is.
 - `userID`: The ID of the user whose status this packet is describing
-- `statuses`: An object with the following fields: `fbAppStatus`, `messengerStatus`, `otherStatus`, `status` and `webStatus`. All can contain any of the following values: `"active"`, `"idle"`, `"invisible"`, `"offline"`.
+<!-- The form of the status request response has changed; it no longer sends five seperate keys - instead it sends a single value. The status states have been condensed to three numbers. -->
+<!-- - `statuses`: An object with the following fields: `fbAppStatus`, `messengerStatus`, `otherStatus`, `status` and `webStatus`. All can contain any of the following values: `"active"`, `"idle"`, `"invisible"`, `"offline"`. -->
```
Schmavery: If you could delete this I'd be super happy to merge :)

lyneca: Delete the commented-out section?

lyneca:
I didn't even see #330 :P 

`presence.a` is just what facebook returns now - I'm not really sure what the fix in #330 actually does.

DOCS.md
Schmavery: Yeah, no point in keeping it around if we're changing it :)

Schmavery:
Haha ok, sounds good.  Facebook changes so fast it can be hard to keep up.

lyneca:
I guess that's the problem with writing an unofficial API for a service that's not built for public access 😛

Schmavery:
Thanks!

Issue #287, opened 2016-06-21
Messages search 


lapinozz:
It would be cool to add message search especially since it look pretty trivial to implement. Here's the post request my browser does: 

The research was for "maison"
`POST /ajax/mercury/search_snippets.php?dpr=1 query=maison&snippetOffset=0&snippetLimit=5&other_user_fbid=100003*******64&identifier=thread_fbid&client=web_messenger`

Something interesting is that you can get message surrounding another one by it's message id:
`POST /ajax/mercury/search_context.php?dpr=1 message_id=mid.1465915683490%3Aefc93467f36f2d3602&limit=16&direction=down&other_user_fbid=100003*******64&client=web_messenger`

This was called when I did the search and then clicked "load more mesages" or something like that
note that it simply changed the `direction` parameter to `up` to get the older message and `down` to get the ewer message

Maybe this could also be helpful for #230

Schmavery:
Sounds cool!
We accept pull requests for new features :smile:

AllanWang:
[This ](https://github.com/AllanWang/AllanBot-Public/blob/master/src/quote.js#L50)is something I made that searches through the messages with the current api features, in case you are interested.

PR #364, opened 2016-12-30
Message search

Mini-Geek:
So this is not very polished or documented, but it does work. This adds the message search feature requested in issue #287.

### DIFF
--- a/src/searchContext.js
+++ b/src/searchContext.js
@@ -3,20 +3,19 @@
 var utils = require("../utils");
 
 module.exports = function(defaultFuncs, api, ctx) {
-  return function searchForMessages(query, threadID, isGroup, snippetOffset, callback) {
+  return function searchContext(messageID, threadID, isGroup, limit, direction, callback) {
     if(!callback && utils.getType(threadID) === 'Function') {
       return callback({error: "Pass a threadID as a second argument."});
     }
     if (!callback) {
-      throw {error: "searchForMessages: need callback"};
+      throw {error: "searchContext: need callback"};
     }
 
     var tmpForm = {
       client: 'web_messenger',
-      query: query,
-      snippetOffset: snippetOffset || 0,
-      snippetLimit: 5,
-      identifier: "thread_fbid",
+      message_id: messageID,
+      limit: limit || 6,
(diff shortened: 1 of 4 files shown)


Schmavery:
Thanks for submitting this @Mini-Geek.  It's a lot of functionality that seems a little complicated to use directly so I'm not sure how to properly expose it, but it's great to have this functionality explored like this and I'll leave the PR open in case anyone else wants to contribute their opinions.

bsansouci:
Hey Thanks for that PR. What does `searchContext` do exactly? It is the same as [getThreadHistory](https://github.com/Schmavery/facebook-chat-api/blob/93b51935781b62605d2c26b8f0957ea674742650/src/getThreadHistory.js)?

Mini-Geek:
When you do a search and click on a result to see the context, search_context.php is called. Also when you click to see more context above or below the found message, it calls search_context.php.

src/searchContext.js
```diff
@@ -0,0 +1,37 @@
+"use strict";
+
+var utils = require("../utils");
+
+module.exports = function(defaultFuncs, api, ctx) {
+  return function searchContext(messageID, threadID, isGroup, limit, direction, callback) {
+    if(!callback && utils.getType(threadID) === 'Function') {
```
bsansouci: Could you do `utils.getType(threadID) !== "String"` instead?

src/searchContext.js
```diff
@@ -0,0 +1,37 @@
...
+      throw {error: "searchContext: need callback"};
+    }
+
+    var tmpForm = {
+      client: 'web_messenger',
+      message_id: messageID,
+      limit: limit || 6,
+      direction: direction,
+    };
+    if (isGroup) {
```
bsansouci: Could you remove `isGroup` and use the same trick we used in `sendMessage` https://github.com/Schmavery/facebook-chat-api/blob/9da530343e2510388bc8cb481f1074d67c463df1/src/sendMessage.js#L155
We make a call to `getUserInfo` and do `Object.keys(res).length > 0`, if it's `true` then it's a 1-1 thread, if it's false then it's a group thread.

PR #369, opened 2017-01-07
Message blocking functions + type: "photo" object fix + listen events fix

bearbear12345:
This function(s) simulate the message blocking features on Messenger.
I'm not entirely sure if the functions are replicating exactly what a browser would do, but the functions are working

May need some tweaking

---

Also updated _formatAttachment() function for photos, #370


src/blockUser.js
```diff
@@ -0,0 +1,37 @@
...
+    defaultFuncs
+      .post("https://www.facebook.com/nfx/block_messages/?thread_fbid=" + userID + "&location=www_chat_head", ctx.jar, {})
+      .then(utils.saveCookies(ctx.jar))
+      .then(utils.parseAndCheckLogin(ctx.jar, defaultFuncs))
+      .then(function(resData) {
+        if (resData.error) {
+          throw resData;
+        }
+        defaultFuncs
+          .post("https://www.facebook.com" + /action="(.+?)"+?/.exec(resData.jsmods.markup[0][1].__html)[1].replace(/&amp;/g, "&"), ctx.jar, {})
```
GAMELASTER: Why writing query in url? Why you don't use `qs` variable instead

bearbear12345: Unless I missed it, there wasn't a developer's documentation for contributing SO I wasn't entirely sure how to do it properly.

I'll give that a try then, cheers!

bearbear12345:
```
function get(url, jar, qs) {
  // I'm still confused about this
  if(getType(qs) === "Object") {
    for(var prop in qs) {
      if(qs.hasOwnProperty(prop) && getType(qs[prop]) === "Object") {
        qs[prop] = JSON.stringify(qs[prop]);
      }
    }
  }
  var op = {
   ...
    url: url,
    method: "GET",
...
```

Looking at [utils.js#L27](../blob/master/utils.js#L27) (the only _qs_ variable I found), that is for a GET request. I was using a POST, since Chrome's developer tools said that it was so

GAMELASTER: `qs` works on POST too, it's a shortcut of QueryString.
From request documentation:
`qs - object containing querystring values to be appended to the uri`

bearbear12345: I won't be able to use ```qs``` though, using ```post(url, jar, form) {...}``` in utils.js#L49

bearbear12345:
^ Also added fix to get the attachment object for photos working, see issue #370

src/blockUser.js
GAMELASTER: Yeah, sorry, I forgot that, forgive me!
@Schmavery maybe will be good to add `qs` argument into `post` function in `utils.js`?

Schmavery: I don't see any problem with adding qs to post :)

utils.js
```diff
@@ -268,18 +268,19 @@ function _formatAttachment(attachment1, attachment2) {
...
-        mimeType: attachment2.mime_type,
-        url: attachment2.image_data.url,
-        width:attachment2.image_data.width,
-        height:attachment2.image_data.height,
+
+        largePreviewUrl: attachment1.large_preview_url,
+        largePreviewWidth: attachment1.large_preview_width,
+        largePreviewHeight: attachment1.large_preview_height,
+        
+        //url: attachment2.image_data.url,
```
Schmavery: Please remove instead of commenting out

Schmavery:
Looks like some great functionality, thanks! 
What do you think about changing the interface to more closely match the pattern of  functions like `changeArchivedStatus`and having one `changeBlockedStatus` or similar that accepts a boolean?
Does this work for group chats or just user IDs?

For the photo fix, what exactly does the resolve function do?  Just get a URL?  I would be inclined to want to just do this automatically upon receiving a photo attachment and then not have to expose the resolve function to users.  Thoughts?

bearbear12345:
re: photo resolving function

Supply a photo ID, and it will return a URL (same as the one when you open Messenger and press download).

---

Yeah sure, I'll make those changes!

bearbear12345:
If I do it automatically, then the events would be delayed until the URL is resolved... isn't that a bad thing?

Schmavery:
Mhh, wouldn't be the first time we make an extra call to give a better user experience.. @bsansouci opinion?

Schmavery:
Great, can you update the docs too?  Also if you could mention that it doesn't work for group threads, that would be great. 
Re: photo attachments, I talked to @bsansouci  and I think we're OK with making the call automatically every time.

bearbear12345:
Yep, sure

Although, I'm not exactly sure how to get the call to work, as the ```_formatMessageAttachment``` isnt a synchronous function, where as the resolving function is called with a promise...

bearbear12345:
Jookesss, this works for me!

Schmavery:
Haha cool, sorry about not getting back to you sooner.  I'll take a look at this ASAP.  If anyone else wants to check that this works for them in the meantime that'd be awesome

Schmavery:
Looks like the readme still needs to be updated.  Can you also make sure the docs for the photo type match the change you're making?  Ideally we make the smallest change to the signature possible to avoid breaking people.

bearbear12345:
There we go!

utils.js
```diff
@@ -399,15 +404,52 @@ function formatMessage(m) {
...
   return {
     type: "event",
-    threadID: m.thread_fbid.toString(),
-    logMessageType: m.log_message_type,
-    logMessageData: m.log_message_data ? Object.assign(m.log_message_data, {
-      removed_participants: (m.log_message_data.removed_participants || []).map(v => v.split("fbid:")[1]),
-    }) : {},
-    logMessageBody: m.log_message_body,
-    author: m.author.split(":")[1]
+    threadID: m.messageMetadata.threadKey.threadFbId,
```
addobandre: When it's a one to one conversation, the threadID is ```m.messageMetadata.threadKey.otherUserFbId```

Schmavery: Thanks for spotting this

bearbear12345: Oh, nice catch! Thanks!

bsansouci:
Looks good to me! Thanks a lot for this PR :)

src/listen.js
```diff
@@ -132,33 +132,45 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+                    case 'change_thread_nickname':
+                    case 'change_thread_icon':
+                      break;
+                    default:
+                      return;
+                  }
+                case 'ThreadName':
+                case 'ParticipantsAddedToGroupThread':
+                case 'ParticipantLeftGroupThread':
+                  return (!ctx.globalOptions.selfListen && formattedEvent.author.toString() === ctx.userID || !ctx.loggedIn) ? undefined : globalCallback(null, utils.formatEvent(v.delta));
```
Schmavery: Looks like we missed a usage of formattedEvent here which is undefined now

Issue #366, opened 2017-01-05
changeThreadColor() callback parameter listed as optional, but errors out if not included


AstroCB:
According to the docs, the `callback` parameter for [`changeThreadColor()`](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#apichangethreadcolorcolor-threadid-callback) is optional, but it throws a whole mess of errors if not included.

Schmavery:
Looks like you're right, thanks for catching this.  Any chance you want to submit a PR it to be properly optional?  We offer lots of emoji hearts in return.

AstroCB:
Sure! #367

PR #373, opened 2017-01-15
Fix emoji callback parameter as optional

AstroCB:


AstroCB:
Callback for `changeThreadEmoji` is optional as per [docs](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#changeThreadColor), but errors out if one isn't provided.

Pretty much the exact same as #366.

Schmavery:
Awesome, thanks!

PR #382, opened 2017-01-31
Add createPoll function

AstroCB:


Schmavery:
Awesome, you did that so fast!
Do you think we can make it so that someone can omit the options but still pass a callback?  I'd imagine you'd have to check if `callback` is null and `options` is of type `function` or something along those lines.

I'm trying to think about whether or not it would make sense to have a different structure for the options.  Tell me if you thought of this already and decided it's crazy.
```
{
  "text from option 1" : true,
  "text from option 2" : false
}
```
Edit: Hmm.. now that I look at it, one clear downside is that it's not ordered like the array would be.  Still would have been a lot nicer haha 😢

AstroCB:
Yeah, that might work a bit better; I'll see what I can do.

Issue #390, opened 2017-02-05
getThreadList return empty names


kamdz:
Hi, I have in my app this code: 
```javascript
app.post('/list', (request, response) => {
  login({appState: JSON.parse(request.body.appState) }, function callback (err, api) {
    if(err) return console.error(err);
    api.getThreadList(0, 10, 'inbox', function(err, arr) {
      if(err) return console.error(err);
      response.send(arr);
    });
  });
});
```
and it's return to me array with 10 elements, but most of them have property: `name: ""`. Only group threads have their names. Single threads with just one person doesn't have anything (even the name of this participant) that I can display on my front. What to do?

kamdz:
Ok, I know where the problem is. Custom nicknames... I will try to fix it and then I will do pull request :)

kamdz:
But, in case that thread is standard (no group, no custom names) it's still doesn't have any specific property to show :/

Schmavery:
Isn't this field supposed to be for the name of the thread, not the name of the participants?  I guess in the case of a 1-person thread we could make this the name of the person you're chatting with, but that seems like more of an open question.  Maybe I'm misunderstanding the problem.

kamdz:
Oh, I see. Check my pr now :)

PR #391, opened 2017-02-05
#390 improve formatThread to not return empty name property

kamdz:
now util function formatThreat return as name property:
-firstly name if it exists, if not then
-custom name of conversation if it set by us, if not then
-other user fbid in one-on-one conversation
-string with coma separated fbids of all users in group conversation

last two options is for example to mapping when we have cached friends list in our map or something like this


Schmavery:
Why not simply have the name empty in the last two cases and have consumers of the API access those ids from the other fields in the returned thread object?

kamdz:
Ok, I see the point, I changed my pr :)

Schmavery:
Great, thanks for doing this, I think I understand better now.
Can we instead expose a nicknames field like returned by getThreadInfo?  An empty thread name is a valid expected return value.

kamdz:
Ok, no problem. Can I return also other_user_fbid value?

kamdz:
Check it now :) 
btw. great project! thanks

Schmavery:
Isn't other_user_fbid the same id as in participants?
I appreciate you sticking with this, I'll give you some insight into why I have to be a little picky.  t's a little tricky to design this api in way that will stay easy to use.  I already think we expose wayyy too many useless fields on this object so I'm trying to get get even more bloat if we can avoid it.

kamdz:
Yes, it's one of them. For example in single conversation participants contains my fbid and other_user_fbid. Ok, I will change it ;p

Schmavery:
Can we just call it "nicknames" to match getThreadInfo.  Also if you could add a quick change to the docs so that people know this will return the new field that would be awesome.

kamdz:
Done!

Schmavery:
Great thanks!

PR #397, opened 2017-02-11
Expose resolvePhotoUrl() for resolving URLs on demand

ivkos:
We can get photo attachments not only through `listen()` where the photo URL resolving was originally implemented, but also through `getThreadHistory()` where the `url` property of photos gets returned as `undefined`. I thought about resolving the URLs there but it didn't seem like a good idea, and would've probably made `getThreadHistory()` significantly slower.

So I think it's a good idea to expose this function to developers, so they can resolve photo URLs whenever they deem necessary.

I've implemented it in `utils.js` so we can reuse the code in `api.listen()` and `api.resolvePhotoUrl()`.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -26,6 +26,7 @@
 * [`api.markAsRead`](#markAsRead)
 * [`api.muteThread`](#muteThread)
 * [`api.removeUserFromGroup`](#removeUserFromGroup)
+* [`api.resolvePhotoUrl`](#resolvePhotoUrl)
 * [`api.searchForThread`](#searchForThread)
 * [`api.sendMessage`](#sendMessage)
 * [`api.sendTypingIndicator`](#sendTypingIndicator)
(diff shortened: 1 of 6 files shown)


bearbear12345:
Hehe looks like we're undoing [this commit](https://github.com/Schmavery/facebook-chat-api/pull/369/commits/c84ad76900045d4656a971708ccb0d0f12c10535)

@Schmavery

Schmavery:
Mhh ok, yes this seems like a valid usecase.
Sorry about that @bearbear12345 :stuck_out_tongue: Thanks for doing the original work on this!

Schmavery:
Can you add a DOCS entry for this function @ivkos ?

ivkos:
Ready to merge, unless it needs any corrections.

Schmavery:
Maybe it's been too long since I wrote this stuff and I'm going crazy, but couldn't we put the logic for resolvePhoto into it's own file, and then just use the api function from the different locations where it's needed without needing to add it to utils.js?  Does that make sense?

ivkos:
Sure, it makes sense. It's most probably possible to do it that way (and certainly better!), so I'll look into that later.

ivkos:
Sorry about the delay. It's ready to merge now.

Schmavery:
Awesome, I think this wasn't exactly what I meant though.  Isn't it possible to just use the `api` here: https://github.com/ivkos/facebook-chat-api/blob/a855d16ab3f334d08f285a24eb287893126f6486/src/listen.js#L10

You should be able to just do `api.resolvePhotoUrl` or whatever with that `api` object.. Really appreciate you helping get this merged ❤️

ivkos:
Underped. 😃

Schmavery:
Thaaaaanks

PR #399, opened 2017-02-12
Make it possible to fetch threads from the 'other' folder

jkp:
This allows users to deal with threads in the "hidden" inbox folder where Facebook sometimes auto routes incoming messages.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -401,7 +401,7 @@ __Arguments__
 
 * `start`: Start index in the list of recently used threads.
 * `end`: End index.
-* `type`: Optional String, can be 'inbox', 'pending', or 'archived'. Inbox is default.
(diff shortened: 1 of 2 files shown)


jkp:
I looked for a test case to update but it seems that there is only rudimentary coverage for this function.  I have tested it locally though and it works as expected.

Schmavery:
Can you add a testcase for the function? :stuck_out_tongue:
If it's too much of a pain I'd imagine it's safe enough to merge but it never hurts to ask ;)

jkp:
Unfortunately I don't think I'll be able to find the time - I'm a newly minted father so finding any hours to code is a now a luxury :)

Schmavery:
Congratulations! 👨 👶 🎉 :1st_place_medal:

PR #400, opened 2017-02-13
Multiple updates and fixes to DOCS and README

ravkr:
* Sort api functions alphabetically
* Add example how to use Github version (a.k.a _Bleeding Edge_) of facebook-chat-api instead of NPM version (_stable_)
* Change examples to use _appState_ instead of _email_+_password_ login
  (it is easier to use, you only have to login once and you don't have to put your password in every source file…)
* Add
  ```js
  const fs = require("fs");
  const login = require("facebook-chat-api");
  ```
  to examples where it was missing
* Some blocks of code were using 2 spaces as indentation - changed to 4 spaces, to match the rest
* change callbacks in examples from these inconsistent forms:
  * `function callback(err, api){}`
  * `function callback (err, api) {}`
  * `function(err, api) {}`
  * `function (err, api) {}`
  (and few other)
  to a single, simpler form:
  * `(err, api) => {}`.
* change `(…, [callback])` to `(…[, callback])`
* many other little fixes. (e.g. `callback(err)` where it should be `console.error(err)`)

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -34,6 +34,35 @@
 
 ---------------------------------------
 
+### Password safety
+
+**Read this** before you _copy+paste_ examples from below.
+
+You should not store Facebook password in your scripts.
+There are few good reasons:
+* People who are standing behind you may look at your "code" and get your password if it is on the screen
+* Backups of source files may be readable by someone else. "_There is nothing secret in my code, why should I ever password protect my backups_"
+* You can't push your code to Github (or any onther service) without removing your password from the file.  Remember: Even if you undo your accidential commit with password, Git doesn't delete it, that commit is just not used but is still readable by everybody.
+* If you change your password in the future (maybe it leaked because _someone_ stored password in source file… oh… well…) you will have to change every occurrence in your scripts
+
+Preferred method is to have `login.js` that saves `AppState` to a file and then use that file from all your scripts.
+This way you can put password in your code for a minute, login to facebook and then remove it.
+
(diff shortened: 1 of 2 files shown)


Schmavery:
Wow thanks for this work @ravkr !  The api really needed a docs overhaul.
The only thing I'm a little reluctant about is the change to using appstate everywhere rather than email+password.  I think the latter is much easier for people to understand/quickly try out.  Otherwise people will first need to know how to generate an appstate (which is not super trivial).  

I definitely don't suggest putting your passwords in source files however.. when we use the api we load a config file with login info or grab it from environment variables.  I had assumed other people would do things like this as well, but I've seen a number of repos on github using our api make this same mistake so maybe I'm wrong.  What do you think about a section near the top of the README talking about password safety or something?  Maybe as an extension to the section about appState.  I'm not sure.

Schmavery:
Sorry about causing conflicts in this.  I'd love to have the pretty new clean docs before publishing the newest version to npm :smile:

Schmavery:
As soon as you can fix this I'll publish 1.3.1 to npm with the new docs 😄

ravkr:
Done :smiley:

Schmavery:
Thanks for doing this so fast!  I'll merge and maybe tweak a couple words when I get a minute, and then your changes will be published for all the world to see :)

ravkr:
I hope that I didn't make many errors :unamused: english is not my native language

Schmavery:
Thanks again @ravkr ! :smile:

PR #401, opened 2017-02-13
Consistent log prefixes and messages

ivkos:
Throughout the library there were log npmlog calls with a missing prefix (i.e. passing the message as the prefix). Fixed that.

Also removed redundant "Error in someFunction" in the prefix/message since this information is present in the function call (e.g. log.error, log.warn)

### DIFF
--- a/src/getThreadList.js
+++ b/src/getThreadList.js
@@ -52,11 +52,11 @@ module.exports = function(defaultFuncs, api, ctx) {
         if (resData.error) {
           throw resData;
         }
-        log.verbose("Response in getThreadList: " + JSON.stringify(resData.payload.threads));
+        log.verbose("getThreadList", JSON.stringify(resData.payload.threads));
         return callback(null, (resData.payload.threads || []).map(utils.formatThread));
       })
       .catch(function(err) {
-        log.error("Error in getThreadList", err);
+        log.error("getThreadList", err);
         return callback(err);
       });
   };
--- a/utils.js
+++ b/utils.js
@@ -575,9 +575,9 @@ function makeDefaults(html, userID) {
 function parseAndCheckLogin(jar, defaultFuncs) {
   return function(data) {
     return bluebird.try(function() {
-      log.verbose("parseAndCheckLogin: " + data.body);
+      log.verbose("parseAndCheckLogin", data.body);
       if (data.statusCode >= 500 && data.statusCode < 600) {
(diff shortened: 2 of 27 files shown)


utils.js
```diff
@@ -575,9 +575,9 @@ function makeDefaults(html, userID) {
 function parseAndCheckLogin(jar, defaultFuncs) {
   return function(data) {
     return bluebird.try(function() {
-      log.verbose("parseAndCheckLogin: " + data.body);
+      log.verbose("parseAndCheckLogin", data.body);
```
ivkos: This may (or may not) as well fix a memory leak I'm currently investigating, as seen here:

ivkos:
Apart from the high memory usage (essentially doubled) caused by v8 storing both:
* `"{ long 700k string }"`
* `"parseAndCheckLogin: { long 700k string }"`

which got fixed by the first commit in the PR, it turned out the `npmlog` module also [stores a record](https://github.com/npm/npmlog/blob/v1.2.1/log.js#L161) of all log messages independent of the logLevel setting. There's an option `log.maxRecordSize` ([defaults to 10000](https://github.com/npm/npmlog/blob/v1.2.1/log.js#L129)). Exposing this to `api.setOptions()` might be useful since storing 10k*700k strings hugely increases memory usage (> 7 GB).

Schmavery:
Wow! Thanks so much for doing this.  We've been aware of some memory issues for a while and never had the chance to diver deeper into them.  The api also sorely needed some logging cleanup too 😄.
❤️ ❤️

Schmavery:
Is this ready to be merged?

ivkos:
Sure! 👍

src/getThreadList.js
```diff
@@ -52,11 +52,11 @@ module.exports = function(defaultFuncs, api, ctx) {
         if (resData.error) {
           throw resData;
         }
-        log.verbose("Response in getThreadList: " + JSON.stringify(resData.payload.threads));
+        log.verbose("getThreadList", JSON.stringify(resData.payload.threads));
```
bsansouci: We could remove this `JSON.stringify` here as we're not concatenating two strings anymore (I'm sure the logger supports printing objs).

bsansouci:
Thanks a loooot for that @ivkos! Here's some hearts for you ❤️ ❤️ ❤️

PR #403, opened 2017-02-15
Fix NaN in listen logs

AstroCB:
Without parentheses around the calculation, string logic evaluates to NaN in the logging from listen.js:
NaN


Schmavery:
Thanks!

PR #412, opened 2017-03-05
Fix unnecessary encoding of options in poll function

AstroCB:


Schmavery:
does this fix a bug?

AstroCB:
Yep
Poll encoding issues

Schmavery:
Awesome thanks.

PR #413, opened 2017-03-05
Adding chatproxy_presence aka buddyList

GAMELASTER:
Returning a list of people (similar to presence) with type of status and timestamp. This event trigger on first pull and in some intervals (server-side, cannot be triggered manually)

BTW. both (chatproxy_presence and buddylist_overlay) returning a field "vc" (in chatproxy_presence).
As far I investigated it have something with status "Active on Messenger"

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -696,6 +696,15 @@ function getType(obj) {
   return Object.prototype.toString.call(obj).slice(8, -1);
 }
 
+function formatProxyPresence(presence, userID) {
+  if(presence.lat === undefined) return null;
+  return {
+    timestamp: presence.lat * 1000,
+    userID: userID,
+    statuses: presence.p === undefined ? 0 : presence.p
+  };
+}
+
 function formatPresence(presence, userID) {
   return {
     type: "presence",
@@ -733,6 +742,7 @@ module.exports = {
   formatMessage: formatMessage,
   formatDeltaMessage: formatDeltaMessage,
   formatEvent: formatEvent,
+  formatProxyPresence: formatProxyPresence,
   formatPresence: formatPresence,
   formatTyp: formatTyp,
   formatCookie: formatCookie,
--- a/utils.js
+++ b/utils.js
@@ -712,6 +712,7 @@ function getType(obj) {
 function formatProxyPresence(presence, userID) {
   if(presence.lat === undefined) return null;
   return {
+    type: "presence",
     timestamp: presence.lat * 1000,
     userID: userID,
     statuses: presence.p === undefined ? 0 : presence.p
--- a/DOCS.md
+++ b/DOCS.md
@@ -652,11 +652,15 @@ Difference between `"read_receipt"` and `"read"`:
 - `"read"` event triggers when the user read other people's messages.
 
 <a name="presence" />
-If enabled through [setOptions](#setOptions), this will also return presence, (`type` will be `"presence"`), which is the online status of the user's friends. The object given to the callback will have the following fields:
+If enabled through [setOptions](#setOptions), this will also return presence or buddyList, (`type` will be `"presence"` or `"buddyList"`), which is the online status of the user's friends. The object given for presence to the callback will have the following fields:
 - `type`: The string "presence".
 - `timestamp`: How old the information is.
 - `userID`: The ID of the user whose status this packet is describing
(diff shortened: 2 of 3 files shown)


utils.js
```diff
@@ -696,6 +696,15 @@ function getType(obj) {
   return Object.prototype.toString.call(obj).slice(8, -1);
 }
 
+function formatProxyPresence(presence, userID) {
+  if(presence.lat === undefined) return null;
+  return {
```
Schmavery: Can we add a `type: "presence" to this to make it match the signature of the other event?

Schmavery:
How is this different than buddylist_overlay?  Will we get duplicate notifications of presence now?

GAMELASTER:
"buddylist_overlay" is event that sending the state changes for friends, for make texts like "Active 5 minutes ago" actual. "chatproxy_presence" is something like left-side bar with all friends and it's activity time. This contains (probably) all friends with their statuses and times :-) .

GAMELASTER:
image
buddy_list overlay still contains only one user and their status update
image
chatproxy_presence with many many entries.

GAMELASTER:
bump

bsansouci:
Hey @GAMELASTER Thanks for doing this!
I think we could merge as soon as you address @Schmavery 's comment on the `utils.js` file :)

utils.js
GAMELASTER: As I written in my reply in pull request, the `type: presence` returns only ONE user presence, this return an list of all users and their presences. For make it match with the signature of default presence event, every user's presence in array will must call callback, what is not good idea.

GAMELASTER:
@bsansouci 
Thanks for reply and sorry, my bad, I replied @Schmavery 's comment in [this](https://github.com/Schmavery/facebook-chat-api/pull/413#issuecomment-284273599) message in PR, but anyway I replied again on his comment.

utils.js
```diff
@@ -696,6 +696,15 @@ function getType(obj) {
...
+    timestamp: presence.lat * 1000,
+    userID: userID,
+    statuses: presence.p === undefined ? 0 : presence.p
```
bsansouci: Can you call this `status` instead.

bsansouci:
Oh I see. Are you still able to get buddylist_overlay nowadays?
Btw could you add documentation that explains that `status: 0` is "not connected" and `status: 2` is connected (it seems from experimentation). I couldn't find any friends with `status: 1` so I don't know what that is lol.

bsansouci:
Oh I just noticed it seems like they both contain timestamps. Do you think we could merge buddy list overlay and proxy-presence? We could transform both to be just `presence`?
If you make the `presence` flow return an array of items, they we could unify them right? So the buddylistoverlay will generally be an array containing just one item and this one will contain a bunch of items. But the information would be the same. What do you think?

bsansouci:
Oh wait, I just noticed that we already document `0` and `2` statuses. So really I think we should unify the current presence and this new one :)

utils.js
GAMELASTER: I leaved it `statuses` because default `presence` object have it as `statuses`

GAMELASTER:
@bsansouci we can unify them, but it will break scripts which working with actual format of presences (so it will broke back-compatilibity). So now I don't know what will be better

GAMELASTER:
So well, how to solve the problem? Change the format of presence or keep back-compatilibity?

Schmavery:
Sorry @GAMELASTER, I thought I responded to this but I guess I must not have.  I think right now it makes sense to iterate through the array and call the callback for each presence object.  If this becomes a performance problem, I think we can address that in the future.  Does that make sense?

GAMELASTER:
@Schmavery I don't think it's performance issue, but it's just not looking good for me, but it will works and keeps back compatilibity. I'm going to change it a commit.

GAMELASTER:
@Schmavery 
Tested, I think it's ready to merge 😉 😊

Schmavery:
Thanks 😄

wayfaredark:
I got a question and is how can i get the user fiend presence info ??? How exactly  is the url structure to perform request!!!!

wayfaredark:
I perform request using https://0-edge-chat.facebook.com/pull, and asend headers to get frend user info but request callback say "INVALID REQUEST"

Issue #415, opened 2017-03-08
Messages with URLs failing to callback


ianw11:
I'm running into issues when messages include urls.  The symptom is that I'm not receiving a callback and I traced the issue all the way back to src/listen.js line 144 in that if messenger catches a link and then shows the preview pane the attach_type is 'share' instead of 'photo'.

For my usage, I don't need anything fancy, I just want the message body.  But I could imagine other people wanting other metadata about this share.

ianw11:
After a little tinkering, I added this after line 151.  It works as far as I can tell

} else {
  globalCallback(null, utils.formatDeltaMessage(v));

Schmavery:
Looks like some people ran into the same problem and used your solution, so thank you!  Do you think you could submit a quick PR with the fix?

Schmavery:
@ianw11 Can you verify that this is fixed by #416?

erdnaxeli:
Seems to be ok, I receive attachments of type "share".

Schmavery:
woo, thanks @erdnaxeli

erdnaxeli:
Maybe could we have a new release now this is fixed? :cat:

GAMELASTER:
@erdnaxeli I think yes

PR #416, opened 2017-03-09
fix: support full share attachments to fix facebook buttons

### LINKED ISSUE #402: Message payload doesn't contain buttons data when there is only one card
Thanks for a great library!

I'm trying on version 1.2.0, to fetch buttons from an attachment that I receive from a bot.  In the attachment (see below) there are 3 buttons but I can only get the title "search engines"

This is partial payload from this library:
```javascript
{ ...
  attachments:
   [ { ...
       type: 'share',
       source: 'Page - testing',
       title: 'search engines',
   } ],
  ...
}
```

This is the attachment:
image

When I'm sniffing facebook chat network I can see that the buttons are part of `call_to_actions` property under an `attachments[0]/mercury/share/target/items[0]` - item with the property `name: "search engines"`

When a bot sends me more than one attachment ("carousel"), I can see the subattachments with relevant details (including buttons)

I'm using the library https://github.com/bluejamesbond/FacebookMessengerBot.js and one of its examples as the bot

noamshemesh:
fix #402 

Not directly related: I think we should break `listen` to several functions to be able to test it better.

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -146,6 +146,8 @@ module.exports = function(defaultFuncs, api, ctx) {
                         if (!err) v.delta.attachments[i].mercury.metadata.url = url;
                         return resolveAttachmentUrl(i + 1);
                       });
+                    } else {
+                      return resolveAttachmentUrl(i + 1);
                     }
                   }
                 })(0)
--- a/utils.js
+++ b/utils.js
@@ -273,7 +273,7 @@ function _formatAttachment(attachment1, attachment2) {
         ID: attachment1.metadata.fbid.toString(),
         filename: fileName,
         thumbnailUrl: attachment1.thumbnail_url,
-        
+
         previewUrl: attachment1.preview_url,
         previewWidth: attachment1.preview_width,
         previewHeight: attachment1.preview_height,
@@ -281,7 +281,7 @@ function _formatAttachment(attachment1, attachment2) {
         largePreviewUrl: attachment1.large_preview_url,
         largePreviewWidth: attachment1.large_preview_width,
         largePreviewHeight: attachment1.large_preview_height,
-        
+
         url: attachment1.metadata.url,
         width: attachment1.metadata.dimensions.split(',')[0],
         height: attachment1.metadata.dimensions.split(',')[1],
@@ -323,6 +323,8 @@ function _formatAttachment(attachment1, attachment2) {
         source: attachment1.share.source,
         title: attachment1.share.title,
         facebookUrl: attachment1.share.uri,
+        target: attachment1.share.target,
+        styleList: attachment1.share.style_list,
         url: attachment2.href,
       };
     case "video":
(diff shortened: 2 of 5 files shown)


Schmavery:
Great, thanks for fixing this.  Is there a reason why we shouldn't check if it's specifically the type of event we're trying to parse?
```
if (v.delta.attachments[i].mercury.attach_type == 'share') {
```

I agree that these switch statements have gotten a little out of hand and should probably be carefully refactored rather than continuing to be patched (turns out one of those is easier to do than the other, which explains the current status).  One first step might be to pull out this inlined resolveAttachmentUrl function.  Might help make it more readable.

I'm not sure I agree with your testing methodology here.  I understand that it's normal to freeze some input fixtures to have a deterministic and fast test, but I'm worried that in so doing, we fail to test the one thing that is most likely to break.  9 times out of 10, when something breaks, it is facebook sending a differently formatted response.  If we're testing against a frozen version, we will miss this bug.

src/listen.js
```diff
@@ -146,6 +146,8 @@ module.exports = function(defaultFuncs, api, ctx) {
                         if (!err) v.delta.attachments[i].mercury.metadata.url = url;
                         return resolveAttachmentUrl(i + 1);
                       });
+                    } else {
+                      return resolveAttachmentUrl(i + 1);
```
Schmavery: Nice catch

utils.js
```diff
@@ -323,6 +323,8 @@ function _formatAttachment(attachment1, attachment2) {
         source: attachment1.share.source,
         title: attachment1.share.title,
         facebookUrl: attachment1.share.uri,
+        target: attachment1.share.target,
+        styleList: attachment1.share.style_list,
```
Schmavery: Can these be documented?

Schmavery:
Is there any change in behaviour after this change for regular attachments?

utils.js
noamshemesh: do you have any documentation example? I'm not sure what you expect :)

noamshemesh:
Shouldn't affect other attachments besides that before adding the `else` block it didn't parse an attachment unless it was a photo (as you've already noticed)

noamshemesh:
Regarding the testing methodology, I totally agree that we must have end to end tests that validate facebook API. 
But I think that we should have more unit tests. The advantages are:
1. we can write them fast
2. they run much much faster
3. we can test internal logic
And 3 is directly related to the reason I had to add this as a unit test. As you mentioned I can't send a button with this library to test the result.

What do you mean by "Does this new code only run when it sees a button"? Can you comment on the specific line?

erdnaxeli:
Does this fix #415 too?

utils.js
Schmavery: https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#apilistencallback

noamshemesh:
@erdnaxeli Yes, I've noticed the same problem and fixed it here: https://github.com/Schmavery/facebook-chat-api/pull/416/files/3c3481129fa13746a5a86fb33f597df4d901ea14#diff-f55ddb040d37ebcc409e74ccbdbdf1a2R150

Schmavery:
Awesome, thanks @noamshemesh, I'll merge as soon as you add the new fields to the docs.

noamshemesh:
Done, sorry for the late response.

Schmavery:
Thanks :)

PR #425, opened 2017-03-24
async support

kabirvirji:
Ran into an issue with async functions since `utils.getType(options) === 'Function'` only checks for regular Functions, and `utils.getType(options)` returns `AsyncFunction` for Async Functions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -423,7 +423,7 @@ function loginHelper(appState, email, password, globalOptions, callback) {
 }
 
 function login(loginData, options, callback) {
-  if(utils.getType(options) === 'Function') {
+  if(utils.getType(options) === 'Function' || utils.getType(options) === 'AsyncFunction') {
     callback = options;
     options = {};
   }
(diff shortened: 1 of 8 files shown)


bsansouci:
Hey @kabirvirji does the API really work when passed an async function?

Schmavery:
I also can't imagine this is the only place that checks if a callback is `=== 'Function'`?

kabirvirji:
@bsansouci I believe it does, that's how I've been using it

@Schmavery The `=== 'Function'` check happens in 8 places, I can add the `AsyncFunction` check to all of them if you think its a good idea, let me know.

bsansouci:
Great to hear! Yeah let's update them in all 8 places to make sure you don't break again :)

kabirvirji:
@bsansouci @Schmavery Done, let me know if anything needs a tweak.

bsansouci:
It looks good! Thanks a lot.

PR #429, opened 2017-03-27
Adding retry limit

GAMELASTER:
@Schmavery 
It's looks fine?

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -602,7 +602,7 @@ function parseAndCheckLogin(jar, defaultFuncs, retryCount) {
         }
         retryCount++;
         var retryTime = Math.floor(Math.random() * 5000);
-        log.warn("parseAndCheckLogin", "Got status code " + data.statusCode + " - " + retryCount + ". attempt to retry in " + retryTime + " seconds...");
+        log.warn("parseAndCheckLogin", "Got status code " + data.statusCode + " - " + retryCount + ". attempt to retry in " + retryTime + " milliseconds...");
         var url = data.request.uri.protocol + "//" + data.request.uri.hostname + data.request.uri.pathname;
         if (data.request.headers['Content-Type'].split(";")[0] === "multipart/form-data") {
           return bluebird
--- a/utils.js
+++ b/utils.js
@@ -593,7 +593,7 @@ function parseAndCheckLogin(jar, defaultFuncs, retryCount) {
     return bluebird.try(function() {
       log.verbose("parseAndCheckLogin", data.body);
       if (data.statusCode >= 500 && data.statusCode < 600) {
-        if (retryCount == 5) {
+        if (retryCount >= 5) {
           throw {
             error: "Request retry failed. Check the `res` and `statusCode` property on this error.",
             statusCode: data.statusCode,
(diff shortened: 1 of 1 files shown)


utils.js
```diff
@@ -585,21 +585,38 @@ function makeDefaults(html, userID) {
...
+            })
+            .then(parseAndCheckLogin(jar, defaultFuncs, retryCount));
         } else {
-          return defaultFuncs
-            .post(url, jar, data.request.formData)
-            .then(parseAndCheckLogin(jar, defaultFuncs));
+          return bluebird
+            .delay(5000)
+            .then(function() {
+              return post(url, jar, data.request.formData);
```
bsansouci: Shouldn't this be ` defaultFuncs.post` instead?

utils.js
```diff
@@ -585,21 +585,38 @@ function makeDefaults(html, userID) {
...
+        }
+        retryCount++;
+        log.warn("parseAndCheckLogin", "Got status code " + data.statusCode + " - " + retryCount + ". attempt to retry in 5 seconds...");
         var url = data.request.uri.protocol + "//" + data.request.uri.hostname + data.request.uri.pathname;
         if (data.request.headers['Content-Type'].split(";")[0] === "multipart/form-data") {
-          return defaultFuncs
-            .postFormData(url, jar, data.request.formData, {})
-            .then(parseAndCheckLogin(jar, defaultFuncs));
+          return bluebird
+            .delay(5000)
```
bsansouci: Could you make it max 5s but random? So create a variable only if the statusCode is `>= 500` and do something like `var retryTime = Math.random() * 4 + 0.5;` or something along those lines.

bsansouci:
Thanks a lot for this PR!

utils.js
Schmavery: Yeah, I think we want to be using this https://github.com/Schmavery/facebook-chat-api/blob/830db449716f49481da8e9b34544b37abe7e0cf3/utils.js#L583 (defaultFuncs.post).  Sorry, this is a little confusing because of the utils.post function (which doesn't add some default values)

utils.js
```diff
@@ -585,21 +585,38 @@ function makeDefaults(html, userID) {
...
+function parseAndCheckLogin(jar, defaultFuncs, retryCount) {
+  if (retryCount == undefined) {
+    retryCount = 0;
+  }
   return function(data) {
     return bluebird.try(function() {
       log.verbose("parseAndCheckLogin", data.body);
       if (data.statusCode >= 500 && data.statusCode < 600) {
-        log.warn("parseAndCheckLogin", "Got status code " + data.statusCode + " retrying...");
+        if (retryCount == 5) {
```
Schmavery: Can you make this if `>= 5` in case something funny ever happens so we don't end up in an infinite loop?

utils.js
GAMELASTER: Yeah, you are right, fixed!

utils.js
GAMELASTER: Yes, I forgot this 😃

GAMELASTER:
It's good now? :blush:

Schmavery:
Looks good, hope this works ✈️

Issue #438, opened 2017-03-31
getThreadHistory message objects senderID's do not have "fbid:" stripped 


mangopearapples:
The senderID's retrieved from messages from getThreadHistory contain the "fbid:" prefix. 
The senderID will look something like: `fbid:100000619620090`
Is this intentional? It is a bit inconsistent from the rest of he api.
Looking at getThreadHistory, the error comes from [v.author](https://github.com/Schmavery/facebook-chat-api/blob/master/src/getThreadHistory.js#L52)

To reproduce:

                            api.getThreadHistory(threadID, 0, 10, undefined, (err, history) => {
					if (err) return console.log(err);

					history.forEach((msg, index) => {
                                            console.log(msg.senderID);
					});
				});

Schmavery:
Fixed by #439

Issue #282, opened 2016-06-04
api can't handle threads that have an alternative ID format


cvigoe:
Most threads have IDs in the form: "1234567890". I've seen a rare thread that has an ID in the form: "id.1234567890". When a thread has this format of ID, any attempt to getFullThread or getUserInfo or any call that relies on a thread ID with this ID format will cause the api to fail. The two screenshots show the difference between the two formats. array[] is an array of threads after a call to getThreadList(). You can see that array[186] has a thread ID in the "normal" form, and array[187] has the rare form. Posting this here in case @Schmavery & co. want to make any changes to accommodate this form of ID and also in case anyone else hasn't noticed this and encounters problems because of it.
<img width="451" alt="screen shot 2016-06-03 at 5 45 48 pm" src="https://cloud.githubusercontent.com/assets/6818920/15796514/d0a4078c-29b3-11e6-8479-a6497b328641.png">
<img width="393" alt="screen shot 2016-06-03 at 5 45 25 pm" src="https://cloud.githubusercontent.com/assets/6818920/15796515/d339795a-29b3-11e6-8aef-ba1aabd4b0bd.png">

Schmavery:
Hmm, we should be stripping off the "id." prefix.  Someone will have to take another look through the getThreadList() code and make sure we're formatting it correctly.  As a workaround for now, manually removing the "id." should result in an id that will work in those calls.

bsansouci:
We "fixed" this issue for messages here https://github.com/Schmavery/facebook-chat-api/blob/f98f3489f73edefb4041213bd3bb94b48b11aed1/utils.js#L375. I think the exact same fix should work for threads. Feel free to submit a PR :)

Schmavery:
This should be fixed by #439.  Reopen if you still have problems :)

PR #439, opened 2017-03-31
Strip 'fbid:' from message senderID in util.js

mangopearapples:
# Fix for [Issue 438](https://github.com/Schmavery/facebook-chat-api/issues/438)

This uses string functions substr and indexOf to strip `fbid:` from senderID as referred in my issue. 

This should be compatible whether the senderID has the `fbid:` prefix or not. If the senderID string does not have the prefix, it will do substr(0) which will not strip anything.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -384,7 +384,7 @@ function formatMessage(m) {
   var obj = {
     type: "message",
     senderName: originalMessage.sender_name,
-    senderID: originalMessage.sender_fbid.toString(),
+    senderID: originalMessage.sender_fbid.toString().substr(originalMessage.sender_fbid.toString().indexOf(':')+1), //Strip "fbid:"
     participantNames: (originalMessage.group_thread_info ? originalMessage.group_thread_info.participant_names : [originalMessage.sender_name.split(' ')[0]]),
     participantIDs: (originalMessage.group_thread_info ? originalMessage.group_thread_info.participant_ids.map(function(v) {return v.toString();}) : [originalMessage.sender_fbid]),
     body: originalMessage.body,
--- a/utils.js
+++ b/utils.js
@@ -369,9 +369,9 @@ function formatDeltaMessage(m){
   var md = m.delta.messageMetadata;
   return {
     type: "message",
-    senderID: md.actorFbId,
+    senderID: formatID(md.actorFbId),
     body: m.delta.body,
-    threadID: (md.threadKey.threadFbId || md.threadKey.otherUserFbId).toString(),
+    threadID: formatID((md.threadKey.threadFbId || md.threadKey.otherUserFbId).toString()),
     messageID: md.messageId,
     attachments: (m.delta.attachments || []).map(v => _formatAttachment(v)),
     timestamp: md.timestamp,
@@ -380,7 +380,11 @@ function formatDeltaMessage(m){
 }
 
 function formatID(id){
-  return id.replace(/(fb)?id[:.]/, "");
+  if(id != undefined && id != null){
+    return id.replace(/(fb)?id[:.]/, "");
+  }else{
+    return id;
+  }
 }
 
 function formatMessage(m) {
@@ -453,7 +457,7 @@ function formatEvent(m) {
 
   return {
     type: "event",
-    threadID: m.messageMetadata.threadKey.threadFbId || m.messageMetadata.threadKey.otherUserFbId,
+    threadID: formatID(m.messageMetadata.threadKey.threadFbId || m.messageMetadata.threadKey.otherUserFbId),
     logMessageType: logMessageType,
     logMessageData: logMessageData,
     logMessageBody: m.messageMetadata.adminText,
@@ -465,7 +469,7 @@ function formatTyp(event) {
   return {
     isTyping: !!event.st,
     from: event.from.toString(),
-    threadID: (event.to || event.thread_fbid || event.from).toString(),
+    threadID: formatID((event.to || event.thread_fbid || event.from).toString()),
     // When receiving typ indication from mobile, `from_mobile` isn't set.
     // If it is, we just use that value.
     fromMobile: event.hasOwnProperty('from_mobile') ? event.from_mobile : true,
@@ -480,7 +484,7 @@ function formatDeltaReadReceipt(delta) {
   return {
     reader: (delta.threadKey.otherUserFbId || delta.actorFbId).toString(),
     time: delta.actionTimestampMs,
-    threadID: (delta.threadKey.otherUserFbId || delta.threadKey.threadFbId).toString(),
+    threadID: formatID((delta.threadKey.otherUserFbId || delta.threadKey.threadFbId).toString()),
     type: 'read_receipt'
   };
 }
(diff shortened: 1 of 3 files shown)


utils.js
```diff
@@ -384,7 +384,7 @@ function formatMessage(m) {
   var obj = {
     type: "message",
     senderName: originalMessage.sender_name,
-    senderID: originalMessage.sender_fbid.toString(),
+    senderID: originalMessage.sender_fbid.toString().substr(originalMessage.sender_fbid.toString().indexOf(':')+1), //Strip "fbid:"
```
GAMELASTER: It's not better to use just `string.replace` function?

GAMELASTER:
Have chat groups too `fbid:` prefix?

mangopearapples:
I thought using indexOf(':') would be better because I saw in a different issue (can't remember which) where the prefix was just "id:". This method would be able to strip the prefix independent of what was before the colon.

Both chat groups and normal conversations have the `fbid:` prefix.

GAMELASTER:
Probably this is world-wide affect, here in Slovakia it's too (we got really late updates).
You already answered, but yes, groups are affected too.

It's possible the delta messages have this prefix too, I will lookup it

GAMELASTER:
It's looks like some functions will be migrating to `fbid` prefix too
image
image

Schmavery:
Could you make sure this very similar issue is also fixed? https://github.com/Schmavery/facebook-chat-api/issues/282

Schmavery:
What about `.replace(/(fb)?id:/, "")`

GAMELASTER:
@Schmavery 
I'm not sure if this all "stripping" is good idea at all.
I think the best idea will be to lookup the Messenger/Facebook sources how they parse "real" threadID, because for sample, Message ID have `mid.` prefix, and without that `mid.` prefix is not possible to delete messages etc. + again Message IDs have many formats. I know it's not threadID related, but in future could be problem with it. 🤔

Schmavery:
@GAMELASTER You could be right, I think there's probably some design decision here.
I'm sure that it's not necessary to have the prefix on the ID, because we can always just add it back on as necessary when making the call.  However, it could allow us to give better error messages if a user passes a message id to a function that is expecting a thread id, for example.

Currently, everything expects ids not to have a prefix, so this is the right fix short term.  If we decide we want to prefix ids, I'd want to do it all at once.  There are a couple edge cases here too, because user ids and thread ids are sometimes treated the same and sometimes not.

Historically, we did it this way because no one cared very much about message ids (you couldn't do many interesting things with them, so usually you didn't need to use it).  Now that we have reactions, which is based on message id, I'd imagine we'll get a couple more messageID-related issues and will have to clean up that area.

@mangopearapples Do you think you could make a utils function that's like "formatID" or something (if we don't already have one) and use that?  If we can gradually move over to having everything use the same function, it will be easier to remove or change behaviour in the future.

mangopearapples:
@Schmavery 
The delimiter in issue #282 is a `.` and not a `:` which could make the regex a bit too complicated. Since `indexOf()` does not use regex, we could use `search` instead? `sender_id.substr(sender_id.search("\\.|:"))` 

Working on util function.

Schmavery:
Is this too complicated?
/(fb)?id[:.]/

mangopearapples:
Yeah I guess not. Thanks.

Participant ID's also started using the `fbid:` prefix since I posted this PR so I have included those in this commit as well.

Schmavery:
@mangopearapples is this ready to merge?  Were you able to add the fix for #282 as well?

mangopearapples:
@Schmavery Yes should be ready unless I've missed out something. I've applied `formatID` to every applicable field (senderID, threadID, participantIDs) I could find.

utils.js
```diff
@@ -369,26 +369,34 @@ function formatDeltaMessage(m){
...
     type: "message",
     senderName: originalMessage.sender_name,
-    senderID: originalMessage.sender_fbid.toString(),
+    senderID: formatID(originalMessage.sender_fbid.toString()),
     participantNames: (originalMessage.group_thread_info ? originalMessage.group_thread_info.participant_names : [originalMessage.sender_name.split(' ')[0]]),
-    participantIDs: (originalMessage.group_thread_info ? originalMessage.group_thread_info.participant_ids.map(function(v) {return v.toString();}) : [originalMessage.sender_fbid]),
+    participantIDs: (originalMessage.group_thread_info ? originalMessage.group_thread_info.participant_ids.map(function(v) {return formatID(v.toString());}) : [formatID(originalMessage.sender_fbid)]),
     body: originalMessage.body,
-    threadID: originalMessage.tid && originalMessage.tid.split(".")[0] === "id" ? originalMessage.tid.split('.')[1] : originalMessage.thread_fbid || originalMessage.other_user_fbid,
+    threadID: formatID(originalMessage.thread_fbid || originalMessage.other_user_fbid),
```
bsansouci: Could you wrap the arguments in parens and add a `.toString()`?
like `threadID:  formatID((originalMessage.thread_fbid || originalMessage.other_user_fbid).toString()),`

utils.js
```diff
@@ -449,7 +457,7 @@ function formatEvent(m) {
 
   return {
     type: "event",
-    threadID: m.messageMetadata.threadKey.threadFbId || m.messageMetadata.threadKey.otherUserFbId,
+    threadID: formatID(m.messageMetadata.threadKey.threadFbId || m.messageMetadata.threadKey.otherUserFbId),
```
bsansouci: Same as above ^

utils.js
```diff
@@ -690,9 +698,9 @@ function formatCookie(arr, url) {
 
 function formatThread(data) {
   return {
-    threadID: data.thread_fbid.toString(),
-    participants: data.participants.map(function(v) { return v.replace('fbid:', ''); }),
-    participantIDs: data.participants.map(function(v) { return v.replace('fbid:', ''); }),
+    threadID: formatID(data.thread_fbid.toString()),
+    participants: data.participants.map(function(v) { return formatID(v.replace('fbid:', '')); }),
```
bsansouci: You can simply do `data.participants.map(formatID)` and same for the line below.

utils.js
```diff
@@ -708,7 +716,7 @@ function formatThread(data) {
     muteSettings: data.muteSettings,
     isCanonicalUser: data.is_canonical_user,
     isCanonical: data.is_canonical,
-    canonicalFbid: data.canonical_fbid,
+    canonicalFbid: formatID(data.canonical_fbid),
```
bsansouci: Could you do `canonicalFbid: formatID(data.canonical_fbid.toString()),`

bsansouci:
Awesome thanks a lot for that! Added some comments, ready to merge once they're addressed :)

bsansouci:
Oh also could you wrap those in `utils.formatID`
https://github.com/Schmavery/facebook-chat-api/blob/9b85a79c3dafca33d6f4d36f6c00fc373903698a/src/getFriendsList.js#L30
and
https://github.com/Schmavery/facebook-chat-api/blob/9b85a79c3dafca33d6f4d36f6c00fc373903698a/src/getUserID.js#L8

utils.js
mangopearapples: Oh wow I really shouldn't have missed that.. I should stop just mindlessly copy pasting lol

utils.js
mangopearapples: @bsansouci The problem with this, is that on `getThreadList`, canonicalFbid is undefined if the thread is not canonical. Calling `toString` on undefined will cause it to throw an exception. Perhaps it would be better to add `toString()` in the `formatID` function?

mangopearapples: Found a way around it using a method similar to the one used on `snippetSender`. Although this means that applications which previously checked whether `canonicalFbid` was undefined would not work as it will now just be an empty string (IMO this is better but may break someones app).

mangopearapples:
All done.

@Schmavery Should be ready to merge now.

bsansouci:
Thanks a lot for this!

Issue #440, opened 2017-03-31
getThreadHistory assumes all returned objects are messages


GEOFBOT:
If there are log entries (such as changing nickname or emoji or color) in the history they are nonetheless treated as messages at https://github.com/Schmavery/facebook-chat-api/blob/9b85a79c3dafca33d6f4d36f6c00fc373903698a/src/getThreadHistory.js#L56, which strips out all of the useful log data.

Schmavery:
This seems related to #313 
I'd love if someone was able to clean up the logic in getThreadHistory, it's pretty old now.  There's been some error handling logic added recently, but it could still use some work.

GEOFBOT:
How about for every message from `getThreadHistory`, instead of calling `utils.formatMessage`, another helper method is called that reads the `action_type` property of the message object and returns the formatted message from the correct formatter? It shouldn't be too difficult or complex; I'll put together a PR shortly. (Thanks for working on this library by the way; it's really really helpful)

Schmavery:
Sounds about right, hopefully that will work!  Let me know if you run into any problems.

Schmavery:
Thanks again @GEOFBOT

Issue #313, opened 2016-08-16
Events are returned as messages with no body while using getThreadHistory


ErmanSayin:
When I use `getThreadHistory`, events are returned with type `message`, and the `body` item is empty.

The event's text is not on the object returned

Schmavery:
Can you print out the result of `resData.payload` [here](https://github.com/Schmavery/facebook-chat-api/blob/master/src/getThreadHistory.js#L39) and see if they've moved this information to a field with a different name?  Could be an easy fix to the formatter if so.

ErmanSayin:
@Schmavery 
Here's two objects, the first is a normal message, and the second is a thread name change event.

```
{ message_id: 'id.1295979970427097',
       threading_id: null,
       offline_threading_id: '6170504581936998170',
       author: 'fbid:100000619620090',
       author_email: '100000****20090@facebook.com',
       ephemeral_ttl_mode: null,
       timestamp: 1471162936057,
       timestamp_absolute: 'Sunday',
       timestamp_relative: 'Sun',
       timestamp_datetime: 'Sunday 10:22am',
       timestamp_time_passed: 1,
       is_unread: true,
       is_forward: false,
       is_filtered_content: false,
       is_filtered_content_bh: false,
       is_filtered_content_account: false,
       is_filtered_content_quasar: false,
       is_filtered_content_invalid_app: false,
       is_sponsored: false,
       commerce_message_type: null,
       customizations: [],
       forward_count: 0,
       forward_message_ids: [],
       source: 'source:messenger:web',
       source_tags: [],
       tags: [ 'source:messenger:web', 'tq', 'cg-enabled', 'inbox' ],
       is_spoof_warning: false,
       folder: 'inbox',
       thread_fbid: '1295979970427097',
       other_user_fbid: null,
       platform_xmd: null,
       message_source: null,
       body: 'test',
       subject: null,
       has_attachment: false,
       attachments: [],
       raw_attachments: null,
       ranges: [],
       meta_ranges: null,
       thread_id: 'id.1295979970427097',
       action_type: 'ma-type:user-generated-message' },
     { message_id: 'mid.1471174770597:eff37047187e9f7b16',
       threading_id: null,
       offline_threading_id: '6170554224725793957',
       author: 'fbid:100000619620090',
       author_email: '100000****20090@facebook.com',
       ephemeral_ttl_mode: null,
       timestamp: 1471174770670,
       timestamp_absolute: 'Sunday',
[...]

PR #441, opened 2017-04-01
getThreadHistory: handle log entry ("event") data properly

GEOFBOT:
#440

### DIFF
--- a/src/getThreadHistory.js
+++ b/src/getThreadHistory.js
@@ -53,7 +53,7 @@ module.exports = function(defaultFuncs, api, ctx) {
               delete v.author;
             });
 
-            callback(null, resData.payload.actions.map(utils.formatMessage));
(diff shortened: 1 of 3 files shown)


Schmavery:
I think the main problem in #313 is that the `type` is `"message"` rather than `"event"` or something, not so much that there's no body.

Schmavery:
@GEOFBOT Can you explain if there's any difference between the events returned by listen and the events returned by getThreadHistory?  Just hoping we can get those as similar as possible.

Schmavery:
You're probably right that making those messages the same is out of scope.  Could you rebase this so that I can merge?

Schmavery:
Going to merge this because I think it's a good step in the right direction.  I still think there's some work that will need to happen to overhaul history messages and make them more similar to listen messages, but that can happen as a separate PR.

Issue #442, opened 2017-04-01
wrong reaction in message_reaction event


ravkr:
`event.reaction` seems to be some weird bytes (eg. `240, 0, 159, 0, 152, 0, 141, 0`) instead of an emoji

test code:
```js
var fs = require("fs");
var login = require("facebook-chat-api");

var example_reaction = "\uD83D\uDE0D" // ":heart_eyes:"
console.log(`Test reaction: ${example_reaction}`);
dump_string(example_reaction);

login({appState: JSON.parse(fs.readFileSync('../appstate.json', 'utf8'))}, (err, api) => {
    if(err) return console.error(err);

    api.setOptions({listenEvents: true, logLevel: "silly"});

    var stopListening = api.listen((err, event) => {
        if(err) return console.error(err);

        if (event.type === "message_reaction") {
            console.log(`Reaction: ${event.reaction}`);
            dump_string(event.reaction);
        }
    });
});

function dump_string(s) {
    var bytes = [];
    for (var i = 0; i < s.length; ++i) {
        var code = s.charCodeAt(i);
        bytes = bytes.concat([code & 0xff, code / 256 >>> 0]);
    }
    console.log('bytes: ' + bytes.join(', '));
}
```

Schmavery:
What's the output of printing it normally?

ravkr:
in the terminal I only see first character
![zrzut ekranu z 2017-04-01 19 51 41](https://cloud.githubusercontent.com/assets/3875202/24581079/aeb8693e-1714-11e7-90f0-e254da3a0f1d.png)

Schmavery:
I wonder if we need something more complicated than [this](https://github.com/Schmavery/facebook-chat-api/blob/c716d06e216d8c372c2c3ae26261d3dcddce6bd1/utils.js#L754) to decode the response.  @GAMELASTER did you come up with this decoder or pull it from the facebook code?  Did you ever see this problem?

GAMELASTER:
@Schmavery It's pulled from Facebook code.
Uh, I browsed the facebook code deeply and found the solution. Reaction requires one more decoding. Sending pull request

Schmavery:
Hey @ravkr did this fix work for you?  I did a quick test on my end and it seemed to help.

bsansouci:
@GAMELASTER @Schmavery You guys are on top of your shit! Nice :D

ravkr:
works perfectly 
thank you :smile:

PR #445, opened 2017-04-01
Implemented Reaction Emoji decoding

GAMELASTER:
Fixing #442 @ravkr could you test it?


Schmavery:
Wow, nice, this seemed to work when I tried it.  Going to merge.

PR #449, opened 2017-04-02
message_reaction: add timestamp

ravkr:


Schmavery:
@GAMELASTER do you have any problem with this?

GAMELASTER:
@Schmavery oh sorry, no no, I just sad because I again forgot something 😞

Schmavery:
Ahh don't worry about it.  I think we have pretty inconsistent handling of timestamps in the other events too.

Issue #262, opened 2016-04-26
GetThreadHistory Parameters


AllanWang:
There's a change I might not have understood the function, but regardless of what I set for "start" and "end", it will always read the messages starting from the most recent ones.

[Here ](https://github.com/AllanWang/AllanBot-Public/blob/5bf3766389b3c7cd8c2992856b29e857ce5df04c/src/quote.js)is my code when I was testing it.

From my understanding, if I set start to 5 and end to 10, it would read the 5th to 10th last messages.
For all my tests, the timestamp was set to Date.now() and I used various numbers for start and end, but they all checked the same messages.

bsansouci:
Try setting timestamp to `null` instead of `Date.now()`. If `timestamp` is set, `start` is effectively ignored.

AllanWang:
I tried that as well and it doesn't seem to change anything.

bsansouci:
Aaah crap. Offset is broken on my side too @Schmavery :(
I'll look more into it tomorrow

alexisohayon:
Same issue here :) How can I fix this on my side?

Schmavery:
This issue has been seen a number of times in different forms.  Add I've said in previous related issues, either Facebook has changed some signature or the docs are wrong. Someone should check again what request the browser is sending to retrieve old messages.

GAMELASTER:
```
messages[user_ids][1000029***48239][offset]:32
messages[user_ids][1000029***48239][timestamp]:1461849806535
messages[user_ids][1000029***48239][limit]:32
```
This works for me nice, where the timestamp is the time of the "oldest message"

bsansouci:
@GAMELASTER can you find out how we could query the messages without a timestamp? I'd be much nicer to query the last 10 messages, and then the 10 messages after that etc... It's pretty likely that they don't allow that anymore because on their front-end they can just take the timestamp of the last message they received when querying for 0-10 and use it for 10-20 right?

Schmavery:
Should be closed by #453?

PR #453, opened 2017-04-04
getThreadHistory parameters adjustments

mangopearapples:
# Addressing #262 

Since `thread_info.php` pretty much ignores the offset data, I thought it would be better to remove that option from the api as it may cause some confusion (as it did to me).

So I changed getThreadHistory to only require `amount` and `timestamp` to get the thread history. I also provided an example of how to load batches of messages in order (like scrolling up on a messenger client) without the need of `offset` functionality. However this means you cannot load messages in history without loading the messages before (more recent) it. Although you couldn't do this before anyhow.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -469,18 +469,43 @@ login({appState: JSON.parse(fs.readFileSync('appstate.json', 'utf8'))}, (err, ap
 ---------------------------------------
 
 <a name="getThreadHistory"></a>
-### api.getThreadHistory(threadID, start, end, timestamp[, callback])
+### api.getThreadHistory(threadID, amount, timestamp[, callback])
 
-Takes a threadID, start and end numbers, a timestamp, and a callback.
+Takes a threadID, amount number, a timestamp, and a callback.
 
 __note__: if you're getting a 500 error, it's possible that you're requesting too many messages. Try reducing that number and see if that works.
 
 __Arguments__
 * `threadID`: A threadID corresponding to the target chat
-* `start`: The ith message in the chat from which to start retrieving history.
-* `end`: The jth message in the chat to which retrieving history.
-* `timestamp`: Used to described the end time. If set, will query messages up to and including `timestamp`.
+* `amount`: The amount of messages to *request*
+* `timestamp`: Used to described the time of the most recent message to load. If timestamp is `undefined`, facebook will load the most recent messages.
 * `callback(error, history)`: If error is null, history will contain an array of message objects.
+
+__Example__
+
+To load 50 messages at a time, we can use `undefined` as the timestamp to retrieve the most recent messages and use the timestamp of the earliest message to load the next 50.
+
+```js
+var timestamp = undefined;
+
+function loadNextThreadHistory(api){
+    api.getThreadHistory(threadID, 50, timestamp, (err, history) => {
+        if(err) return console.error(err);
+
+        /*
+            Since the timestamp is from a previous loaded message, 
+            that message will be included in this history so we can discard it unless it is the first load.
+        */
+        if(timestamp != undefined) history.pop();
+
+        /*
+            Handle message history
+        */
+
+        timestamp = history[0].timestamp;
+    });
+}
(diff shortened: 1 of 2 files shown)


bsansouci:
Awesome! Thanks a lot for fixing this.
The example is a good choice.

ravkr:
What? When it started ignoring end? I was able to pull all messages from thread using only start and end about a month ago :/

mangopearapples:
@ravkr Can you try it again with the old `getThreadHistory`? It's still ignoring `offset` for me.

```js
api.getThreadHistory(threadID, 5, 10, undefined, (err, history)=>{
	history.forEach((msg, index)=>{
		console.log(msg.body);
	});
	app.quit();
});
```

The code above basically ignores the `5` and gets me the 5 most recent messages instead of the 5 before that.

GAMELASTER:
If I remember good, the `getThreadPictures` have same behaviour

ravkr:
yeah... `offset` seems not to work anymore :(

BilyanHadzhi:
Unfortunately, this doesn't seem to work on my side.

`timestamp` is treated as the timestamp of the oldest message so it just loads all messages after it.
So I'm just stuck loading the same messages even though I'm updating `timestamp`.

ravkr:
@bilyanHadzhi can you show us your code?

BilyanHadzhi:
@ravkr Sure.

I'm trying to export all messages from a huge group chat (>32,000 messages) to a JSON file.

```js
const login = require('facebook-chat-api');
const jsonfile = require('jsonfile');

const credentials = {
  email: process.env.FB_EMAIL,
  password: process.env.FB_PASSWORD,
};

const file = '/home/billy/Coding/projects/z-class-chat-analysis/src/messages.json';
const threadId = 814129402005312;

let timestamp = undefined;
let allMessages = [];

const writeNextHistory = api => {
  api.getThreadHistory(threadId, 3, timestamp, (err, history) => {
    if (err) {
      console.error(err);
    }

    if (timestamp !== undefined) {
      history.pop();
    }

    console.log('timestamp', timestamp);
    history.forEach(message => console.log(`${message.body}: timestamp is ${message.timestamp}`));
    console.log();

    allMessages = history.concat(allMessages);
    timestamp = history[0].timestamp;
  });
};

login(credentials, (err, api) => {
  if (err) {
    console.error(err);
  }

  writeNextHistory(api);
  writeNextHistory(api);
});

```

The result is a little something like this: 
```
timestamp undefined
Thanks: timestamp is 1493019246820
You're welcome! Cya later: timestamp is 1493019392190
Cya: timestamp is 1493019397398

timestamp 1493019246820
Thanks: timestamp is 1493019246820
You're welcome! Cya later: timestamp is 1493019392190
```

As you can see, it's using the message with timestamp of `timestamp` as the oldest, not the most recent message.

mangopearapples:
@BilyanHadzhi The API works asynchronously. So the second `writeNextHistory(api)` is called before the first api call has even started. You will need to implement a callback in your `writeNextHistory` function to execute the callback after `timestamp = history[0].timestamp`.

In effect, you will have something like this:

```Javascript
login(credentials, (err, api) => {
  if (err) {
    console.error(err);
  }

  writeNextHistory(api, () => writeNextHistory(api));
});
```

ravkr:
you are calling `writeNextHistory` twice, without waiting for result
if you try this:
```js
// ...
const writeNextHistory = api => {
    console.log('timestamp', timestamp);
    api.getThreadHistory(threadId, 3, timestamp, (err, history) => {
// ...
```
you will see that both calls are made when `timestamp` is `undefined`

your second call to `writeNextHistory` should be made **after** receiving response from the first call

BilyanHadzhi:
Ah, I see now.

So, I guess if I wanted to call this function again and again till I've loaded some arbitrary number of messages (and stored them in a JSON file in the process), I guess I'd write some sort of recursive callback?

Thank you very much again, guys, I'll ask again if need be.

PR #457, opened 2017-04-06
sendTypingIndicator: make callback optional

ravkr:
I was getting this error:

```
Unhandled rejection TypeError: callback is not a function
    at /var/www/node/messenger/node_modules/facebook-chat-api/src/sendTypingIndicator.js:39:18
    at tryCatcher (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:510:31)
    at Promise._settlePromiseAt (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:584:18)
    at Promise._settlePromises (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:700:14)
    at Async._drainQueue (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/var/www/node/messenger/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

when using:
```js
end = api.sendTypingIndicator(event.threadID, function() {});
// run some background process
api.sendMessage(msg, event.threadID, end);
```

and it turned out that `function end(cb)` was called with `cb` being an object instead of a function.

### DIFF
--- a/src/sendTypingIndicator.js
+++ b/src/sendTypingIndicator.js
@@ -12,7 +12,7 @@ module.exports = function(defaultFuncs, api, ctx) {
       thread: threadID
     };
 
-    // Check if thread is single person chat or group chat
+    // Check if thread is a single person chat or a group chat
     // More info on this is in api.sendMessage
     api.getUserInfo(threadID, function(err, res) {
       if (err) {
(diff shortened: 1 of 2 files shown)


Schmavery:
Is this behaviour consistent with code you're seeing elsewhere in the api? I don't think I understand why we would want to allow people to pass in objects or other types that get silently swapped for an empty callback.
Do you think you could change this so that we explicitly fail when they pass something in that isn't a function?  That seems to make more sense to me, if anything.

Schmavery:
Ahh sorry, I reread the example and see what you were trying to do.  Not sure what the right option is here.  More of a design question at this point.

ravkr:
how about adding some warning when swapping for an empty callback?

Schmavery:
Ok sure.  I talked to @bsansouci and he thought this is a usecase worth supporting.  Let's go with the warning rather than failing.

ravkr:
Is everything okey now?

src/sendTypingIndicator.js
```diff
@@ -42,15 +42,21 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+      callback = () => {};
     }
 
     makeTypingIndicator(true, threadID, callback);
 
-    // TODO: document that we return the stop/cancel functions now
     return function end(cb) {
-      makeTypingIndicator(false, threadID, cb || function() {});
+      if(!callback || (utils.getType(threadID) !== 'Function' && utils.getType(threadID) !== 'AsyncFunction')) {
+        cb = () => {};
```
Schmavery: Warning here like above?

ravkr: oh god... I should get more sleep :laughing:

Schmavery:
Thanks.

PR #460, opened 2017-04-09
Added Mentions

MikeShi42:
Added the ability to use mentions in the sendMessage object so that a bot can ping specific people. 

Main issue is that tag location is determined with `indexOf` which is a bit scrappy. It provides a good balance of ease of use as the tag text can just be passed in, instead of the user having to figure out where the tag is in the string themselves. Can be changed to also explicitly take in beginning + ending index of the string as well.

### DIFF
--- a/src/sendMessage.js
+++ b/src/sendMessage.js
@@ -11,6 +11,7 @@ var allowedProperties = {
   emoji: true,
   emojiSize: true,
   body: true,
+  mentions: true,
 };
 
 module.exports = function(defaultFuncs, api, ctx) {
@@ -236,6 +237,20 @@ module.exports = function(defaultFuncs, api, ctx) {
     }
   }
 
+  function handleMention(msg, form, callback, cb) {
+    if (msg.mentions) {
+        for (let i=0; i < msg.mentions.length; i++) {
+            let mention = msg.mentions[i];
+            let tag = mention.tag || '';
+            let id = mention.id || 0;
+            form['profile_xmd[' + i + '][offset]'] = msg.body.indexOf(tag);
+            form['profile_xmd[' + i + '][length]'] = tag.length;
+            form['profile_xmd[' + i + '][id]'] = id;
+        }
+    }
+    cb();
+  }
+
   return function sendMessage(msg, threadID, callback) {
     if(!callback && (utils.getType(threadID) === 'Function' || utils.getType(threadID) === 'AsyncFunction')) {
       return callback({error: "Pass a threadID as a second argument."});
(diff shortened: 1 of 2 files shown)


Schmavery:
Awesome work @MikeShi42 
What do people think about an optional `fromIndex` like is offered by javascript's `indexOf `operator?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf

It could default to 0 and that way people still have all the power they need just in case.

Schmavery:
Also: Is there any verification of input we should be doing?  i.e. checking if `msg.body.indexOf(tag)` is `-1` and if so calling the callback with an error?  I'd probably prefer this to returning some generic facebook error.

MikeShi42:
@Schmavery we can def do some client side checking as well (sorry the code is hackish). It seems like FB will happily take the invalid input and just toss it out when rendering in terms of errors, it would be nice for the library to tell the user that their mention is 'invalid'. As for the fromIndex, are you thinking of it being an optional property in the passed in object? It seems like a good idea, and along that lines we can also include a toIndex for more flexibility.

Additionally, how do you feel about the documentation? I know the API I've created is a bit bulkier than other fields so I had some trouble trying to explain it well and just attached an example which hopefully would clear things up.

Schmavery:
I think the example works well for docs.  You can make a note in the description of the field to see below for an example (bonus points for making it link).

I think it makes sense to just add the `fromIndex` (`toIndex` is not really necessary and doesn't match the signature of `indexOf`).  Adding it as an optional field in the passed in objects (default = 0) would be great.

Yeah basically some error or warning when `msg.body.indexOf(tag, fromIndex) < 0` or something would be great.

src/sendMessage.js
```diff
@@ -239,14 +239,27 @@ module.exports = function(defaultFuncs, api, ctx) {
...
-            let id = mention.id || 0;
-            form['profile_xmd[' + i + '][offset]'] = msg.body.indexOf(tag);
-            form['profile_xmd[' + i + '][length]'] = tag.length;
-            form['profile_xmd[' + i + '][id]'] = id;
-        }
+      for (let i=0; i < msg.mentions.length; i++) {
+        const mention = msg.mentions[i];
+
+        const tag = mention.tag;
+        if (typeof tag !== "string")
```
Schmavery: can you add curly braces :D

MikeShi42: yup c:

Schmavery:
Thanks @MikeShi42 !

PR #462, opened 2017-04-10
Allow the first result for a search to be a non user and include the type in the returned data

mitchcapper:
Discussion in #459

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,3 +227,4 @@ login({appState: JSON.parse(fs.readFileSync('appstate.json', 'utf8'))}, (err, ap
 - [Klink](https://github.com/KeNt178/klink) - This Chrome extension will 1-click share the link of your active tab over Facebook Messenger
 - [Botyo](https://github.com/ivkos/botyo) - Modular bot designed for group chat rooms on Facebook
 - [matrix-puppet-facebook](https://github.com/matrix-hacks/matrix-puppet-facebook) - A facebook bridge for [matrix](https://matrix.org)
+- [facebot](https://github.com/Weetbix/facebot) - A facebook bridge for Slack.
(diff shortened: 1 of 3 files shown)


Schmavery:
Can you document the `type` field, with maybe a couple of examples of common values?

mitchcapper:
Done. There was no documentation for the array at all, so expanded it.  Note getuserinfo returns the same type variable let me know if you want the options noted there as well.  Note I didn't find an official list of types, so just ran a bunch of queries to find what I could.

README.md
```diff
@@ -227,3 +227,4 @@ login({appState: JSON.parse(fs.readFileSync('appstate.json', 'utf8'))}, (err, ap
 - [Klink](https://github.com/KeNt178/klink) - This Chrome extension will 1-click share the link of your active tab over Facebook Messenger
 - [Botyo](https://github.com/ivkos/botyo) - Modular bot designed for group chat rooms on Facebook
 - [matrix-puppet-facebook](https://github.com/matrix-hacks/matrix-puppet-facebook) - A facebook bridge for [matrix](https://matrix.org)
+- [facebot](https://github.com/Weetbix/facebot) - A facebook bridge for Slack.
```
Schmavery: Some merge problem?

Schmavery:
Great, yeah anywhere we change the object should be noted (i.e. getUserInfo if applicable)

mitchcapper:
I didn't change getUserInfo it already had the "type" property it just doesn't document what type is or any of the other properties it has.

I did add a reference to facebot in this commit, do you want a separate PR for that or just not at all?

Schmavery:
Ah ok.  Can you add it then? haha 😛 

Ok.  We can probably just leave the facebook bit in, next time separate PR would be even better, but I won't tell anyone haha

mitchcapper:
done.

Issue #452, opened 2017-04-03
Add option to turn off extra buddylist presence events


danielriddell21:
My FB Bot hasn't been up to any spammy activities. Ever since I added the presence to the set options, I recieve alot of console log about it showing all the people every couple of mins. Before It didnt do this it would only display when someone goes on of off. Never in a bulk. I am not sure whats happened. Since i updated it seems to do all this spammy console log stuff and It gets irritating. Is this due to your code or by a change by facebook. 

![image](https://cloud.githubusercontent.com/assets/25375432/24613498/67f903a6-1880-11e7-9935-cf07e5df5471.png)

bobland1:
I am going to wait for the temp ban to be lifted and stop the presence calling for now. But for some reason ever since updating. I received a lot of spammy console log messages from presence.

Schmavery:
I think there are 2 separate issues here.  The "presence spam" is as a result of https://github.com/Schmavery/facebook-chat-api/pull/413.  It's possible that we could add an option to turn this off or change the log priority, but remember that you can adjust the global logging level whenever you want.

Being temporarily banned is definitely not fun and we'd like to do all we can to avoid that happening.  Unfortunately it's also very hard to diagnose.  Please let us know if there's some idea you have of things people should avoid doing to avoid triggering this.

Schmavery:
I have a hard time thinking that the new presence notifications are the cause of this, because I'm pretty sure we were already receiving this information before and throwing it away, and there's no extra calls that we being made.  @GAMELASTER can confirm.

@bobland1 what other api functionalities were you using?

bobland1:
Aight thank you very much :) And, ill send you my code privately so you can see :) I have commented some stuff out so i dont get temp banned again.

bobland1:
![image](https://cloud.githubusercontent.com/assets/25375432/24621909/18368aee-189b-11e7-874c-3351f7933f74.png)
![image](https://cloud.githubusercontent.com/assets/25375432/24621968/42e327b6-189b-11e7-9ada-30f99c27cf64.png)
Is why i got banned. I use this
![image](https://cloud.githubusercontent.com/assets/25375432/24622001/5a670a1a-189b-11e7-918b-e1c2a87dbbb7.png)

bobland1:
But due to the spam at the start, i feel that is what caused the ban. Is there a way i am able to turn that off, so i dont receive the spam at the start?

GAMELASTER:
Heya,
[...]

PR #464, opened 2017-04-10
Adding buddy list Filtering

danielriddell21:
#452 
Buddy List filtering as it is very noisy and Spammy at the moment

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -121,7 +121,7 @@ module.exports = function(defaultFuncs, api, ctx) {
               break;
             case 'chatproxy-presence':
               // TODO: what happens when you're logged in as a page?
-              if(!ctx.globalOptions.updatePresence) {
+              if(!ctx.globalOptions.batchPresence || !ctx.globalOptions.updatePresence) {
                 return;
               }
 
--- a/src/listen.js
+++ b/src/listen.js
@@ -121,7 +121,7 @@ module.exports = function(defaultFuncs, api, ctx) {
               break;
             case 'chatproxy-presence':
               // TODO: what happens when you're logged in as a page?
-              if(!ctx.globalOptions.updatePresence || !ctx.globalOptions.updateBuddylist) {
+              if(!ctx.globalOptions.updatePresence) {
                 return;
               }
 
@@ -138,7 +138,7 @@ module.exports = function(defaultFuncs, api, ctx) {
               break;
             case 'buddylist_overlay':
               // TODO: what happens when you're logged in as a page?
-              if(!ctx.globalOptions.updatePresence) {
+              if(!ctx.globalOptions.updateBuddylist) {
                 return;
               }
               // There should be only one key inside overlay
--- a/src/listen.js
+++ b/src/listen.js
@@ -138,7 +138,7 @@ module.exports = function(defaultFuncs, api, ctx) {
               break;
             case 'buddylist_overlay':
               // TODO: what happens when you're logged in as a page?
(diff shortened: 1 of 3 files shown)


bobland1:
I have started work on this. Very minor as of now but I'll look more into this. Hopefully I'll manage to work something out.

Schmavery:
@bobland1 Looks like you're making these changes though the web interface, so I just wanted to check if you know that it's possible to sync the repo to your local machine.
When you do this, then you can hook up whatever code you have that's using the api to use your local version instead fairly easily, and in that way, test if it's working.
Feel free to ignore if you already know all this 😛

src/listen.js
```diff
@@ -121,21 +121,27 @@ module.exports = function(defaultFuncs, api, ctx) {
...
-                for(var userID in v.buddyList) {
-                  var formattedPresence = utils.formatProxyPresence(v.buddyList[userID], userID);
-                  if(formattedPresence != null)
-                  {
-                    globalCallback(null, formattedPresence);
-                  }
-                }
-                return;
-              }
+              case 'buddylist':
```
Schmavery: Not sure what's happening here, I think maybe you misunderstood me before.
You should be able to simply replace
```
if(!ctx.globalOptions.updatePresence) {
```

with something like 

```
if(!ctx.globalOptions.updatePresence || ctx.globalOptions.updateBuddyList) {
```

bobland1:
Yeah, I'm updating through the web interface due to being on my phone then. In the morning when I'm on my PC I can do all the pushing and syncing

bobland1:
And yeah I seem to have misunderstood 😂

Schmavery:
I've been trying to think about how to offer the best api here and I think we should tweak this design a little.

Let's call it `batchPresence` and have it turned off by default.  Then we can have it so that if you have `updatePresence` *and* `batchPresence` turned on, you will get the new events.

The reason we will call it `batchPresence` is that now that this is behind a flag (so it's not a breaking change) we can update the signature of the presence object to return an array if there are multiple elements.

@GAMELASTER sorry for the back and forth on the signature here.  Any chance you can help @bobland1 with this change?

bobland1:
image

src/listen.js
```diff
@@ -135,11 +135,10 @@ module.exports = function(defaultFuncs, api, ctx) {
                 }
                 return;
               }
-
               break;
             case 'buddylist_overlay':
               // TODO: what happens when you're logged in as a page?
-              if(!ctx.globalOptions.updatePresence) {
+              if(!ctx.globalOptions.batchPresence) {
```
GAMELASTER: Make here to if the updatePresence and batchPresence must be a true

bobland1: Okay? I'll change that. Little confused as to why but will change now

bobland1: Added minor adjustments and fixed as you correctly pointed out updatePresence. Told ya I was prone to typos 😄

GAMELASTER:
Well, this looks well and okay! :blush:

bobland1:
Thank you :)

src/listen.js
Qualitymix: I'm having trouble keeping a persistent friend list that is accurate.  Every 2 minutes or so, Facebook will tell me a whole bunch of people who are actually online, are now offline.  I'm assuming in this case the 0 means their status hasn't changed.     This is code fix that?

Qualitymix: so how can i properly use batchPresence?

PR #475, opened 2017-04-27
Fix inconsistencies between thread data in getThreadInfo and getThreadList

mangopearapples:
# Fixing thread data inconsistencies 

The data you get from threads in `getThreadInfo` and `getThreadList` are wildly different. This PR adds the data in `getThreadInfo` that wasn't into `utils.formatThread` and applies `utils.formatThread` to the thread in `getThreadInfo` so that you could get the same data you get from threads in `getThreadList` in `getThreadInfo`.


ravkr:
oh god... i started doing this few days ago :/

Schmavery:
Just to confirm, this doesn't remove/rename any fields and doesn't make any more requests to facebook than before, right?
If that's the case, this looks like a good idea 😄

mangopearapples:
Yes totally backwards compatible (although this is by chance :P) as `utils.formatThread` had most of the properties from `getThreadInfo` anyway minus a few so I just added the properties `utils.formatThread` was missing and applied it to `getThreadInfo`.

The actual requests to facebook haven't been touched at all.

Schmavery:
Thanks again @mangopearapples

Issue #463, opened 2017-04-10
Refactor listen docs


Schmavery:
The docs for listen have gotten basically unreadable with the recent addition of several new types of message (no fault to anyone, thanks everybody for the contributions! 😄 ).  It would be great if these could be rewritten to be easier to read.  I'm imagining some tabular structure, maybe, with attempts to describe the fields where possible.
We might need a couple of tables, one to handle the types of messages (..this might just contain one thing?), one for the types of events and one for the types of attachments.  I feel like most solutions would be better than what we have now.

Unfortunately github markdown doesn't support lists inside of tables, but you can hack it ok with the `<br>` element.  Definitely open to suggestion here.  How do other apis solve this problem?

```
|  Event Type   |     Fields    |
| ------------- | ------------- |
| `"event"`  | `threadID`: The thread which the event was performed in.  <br> `author`: The person who performed the event. <br> ...|
| `"typ"`  | `threadID`: The thread which the event was performed in.  <br> `author`: The person who performed the event. <br> ... |
```

| Event Type  | Fields |
| ------------- | ------------- |
| `"event"`  | `threadID`: The thread which the event was performed in.  <br> `author`: The person who performed the event. <br> ...|
| `"typ"`  | `threadID`: The thread which the event was performed in.  <br> `author`: The person who performed the event. <br> ... |

AllanWang:
I'll take a look at this after finals

ravkr:
I will have some time after April 22nd so I'll try to help you guys :smile:

PR #476, opened 2017-04-28
updating getThreadList

ravkr:
*You don't need to review changes yet, **it is not finished**. I just wanted to mark that I started this so we can discuss it*

---

`api.getThreadList` uses many non-existent properties and also doesn't have many new informations that facebook sends in their response

currently I'm doing all my tests in `getThreadListEx.js`

---

I couldn't find these in any facebook responses:
`thread.former_participants`
`thread.snippet_has_attachment`
`thread.last_message_id`
`thread.blocked_participants`
`thread.root_message_threading_id`
so I removed them

this is new but I am not sure if this will be object or string:
`thread.cannot_reply_reason`

---

the old version was only returning formatted `threads` but it would be very useful to get all info that facebook sent to us, especially `participants` (why would we have to call `getUserInfo()` for every user to get their info while it was already sent to us?)

```json
{
    "__ar": 1,
    "__sf": "iw",
    "payload": {
        "threads": [ ... ],
        "ordered_threadlists": [ ... ],
        "participants": [ ... ],
        "unseen_thread_fbids": [ ... ],
        "roger": { ... },
        "delivery_receipts": [ ... ],
        "payload_source": "server_fetch_threadlist_info"
    },
    "bootloadable": {},
    "ixData": {},
    "lid": "6415863198447271708"
}
```

---

I will document everything after all the rest is done and I'll try to do it in a way suggested in #463


Schmavery:
Thanks for looking into this.  If fields truly are always empty, I think we can remove them without being scared about backwards compatibility.
Lets be careful about what we add to make sure it's useful and consistent.  I don't want to simply return the object from facebook, as that changes frequently and breaks people.

ravkr:
but I was thinking about returning object with arrays (threads, participants, …) instead of just array of threads

Schmavery:
Can you give a concrete example of the object you would want to return so that we can discuss it?

ravkr:
currently I was thinking about something like that:
```js
{
    threads: [], // just updated version of what was before
    participants: [], // all users participating in threads above
    roger: [] // probably we need another name... here will be list of these "seen" dots
}
```

Schmavery:
What data (specifically) is contained in participants/roger?  Can we insert the data for participants and roger into their relevant threads?

ravkr:
yes, it would be possible to insert users' data into threads... but we would have to duplicate it
if one person is in 10 different threads then we would have to copy that user's info to every thread
it doesn't sounds good :cry: 

`threadid` and `userid` are just numbers (`000000000000000`)
```json
"participants": [
    {
        "id": "fbid:000000000000000",
        "fbid": "000000000000000",
        "gender": 1,
        "href": "https://www.facebook.com/profile.php?id=000000000000000",
        "image_src": "https://scontent-amt2-1.xx.fbcdn.net/..............",
        "big_image_src": "https://scontent-amt2-1.xx.fbcdn.net/..............",
        "name": ".....",
        "short_name": ".....",
        "employee": false,
        "is_employee_away": false,
        "is_user_focusing": false,
        "type": "user",
        "vanity": "",
        "accepts_messenger_user_feedback": false,
        "is_friend": true,
        "is_messenger_user": true,
        "is_business_enabled": false,
        "is_verified": false,
        "is_messenger_platform_bot": false,
        "timezone": null,
        "work_id": "",
        "personal_id": "",
        "is_messenger_blocked": false
    },
],
"roger": {
    "threadid (here is same as userid - its private conversation)": {
        "userid": {
            "watermark": 1493114634862,
            "action": 1493115564400
        }
    },
    "another threadid": {
        "1st iserid": {
            "watermark": 1493027499150,
            "action": 1493028335556
        },
        "2nd userid": {
            "watermark": 1493113521339,
            "action": 1493113613892
        },
        "3rd userid": {
            "watermark": 1493115481223,
            "action": 1493117237603
        }
    }
}
```

Schmavery:
hey @ravkr, what's the status of this?

ravkr:
Well... I didn't have much time :/ sorry 😢 
Probably we would have to analyze this again (something could have changed)

Issue #467, opened 2017-04-15
Messenger emoji image URL resolver function


GEOFBOT:
I've been trying to make a log exporter tool, and I needed to get the image URLs for Messenger-style emojis. I think I've figured out how the URLs are generated, and I was wondering if a function to get emoji image URLs (similar in nature, somewhat, to `resolvePhotoUrl`) would be a relevant addition to this library.

ravkr:
I think it would be great

Schmavery:
This has been merged.

PR #477, opened 2017-04-30
Utility function for getting the image URL for a Messenger-style emoji

GEOFBOT:
#467 

Example usage:
`utils.getEmojiUrl('🐍', 128)` returns `https://static.xx.fbcdn.net/images/emoji.php/v8/ze1/1.5/128/1f40d.png` which is the emoji asset url for the snake emoji:

### DIFF
--- a/src/getEmojiUrl.js
+++ b/src/getEmojiUrl.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const utils = require("../utils");
+const util = require("util");
 
 module.exports = function() {
   return function getEmojiUrl(c, size, pixelRatio) {
--- a/src/getEmojiUrl.js
+++ b/src/getEmojiUrl.js
@@ -1,0 +1,24 @@
+"use strict";
+
+const utils = require("../utils");
+
+module.exports = function() {
+  return function getEmojiUrl(c, size, pixelRatio) {
+    /*
+     Resolves Facebook Messenger emoji image asset URL for an emoji character.
+     Supported sizes are 32, 64, and 128.
+     Supported pixel ratios are '1.0' and '1.5' (possibly more; haven't tested)
+     */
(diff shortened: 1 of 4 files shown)


ravkr:
I think this should be in a separate file and added to https://github.com/Schmavery/facebook-chat-api/blob/master/index.js#L75
(and also to DOCS)

Schmavery:
@ravkr is right, if this function is meant for general users (not just internally), then it should be exposed in that way.  Let me know if you need help with that.
Is this a feature that anyone thinks they would use?  Trying to understand the usecase.

Schmavery:
Ok, seems reasonable.  Feel free to add a link to your tool to the end of the README  (in a separate PR) when it's ready 😄

Schmavery:
Can we assume for now that everyone using the api is logging in using it?  I think I'd rather go with that for now.

src/getEmojiUrl.js
```diff
@@ -0,0 +1,24 @@
...
+  return function getEmojiUrl(c, size, pixelRatio) {
+    /*
+     Resolves Facebook Messenger emoji image asset URL for an emoji character.
+     Supported sizes are 32, 64, and 128.
+     Supported pixel ratios are '1.0' and '1.5' (possibly more; haven't tested)
+     */
+    const baseUrl = 'https://static.xx.fbcdn.net/images/emoji.php/v8/z%s/%s';
+    pixelRatio = pixelRatio || '1.0';
+
+    let ending = util.format('%s/%s/%s.png', pixelRatio, size, c.codePointAt(0).toString(16));
```
Schmavery: Does util not need to be 'require'd?

GEOFBOT: For some reason it worked when I tested it, but I've added the proper require

Schmavery:
Thansk @GEOFBOT :)

Issue #480, opened 2017-05-07
Create changelog


AllanWang:
I think it would be worthwhile to continue creating releases on the github repo (latest is 1.2.0) and add either a Changelog.md or changelog in the releases so users can see what has changed. Features are added all the time, and some of them may break bot code for older versions, and it would be nice to have an easier way to distinguish which portions need to be changed without sifting through the commits.

ravkr:
Great idea! I was thinking about it few weeks ago... :/
And I think that we should also put some clues how to update code (what someone have to change to get the same result as before)

Schmavery:
@AllanWang I suppose we could start this, if someone wants to take a quick look at the past couple versions and see what has changed.  I've held off up to this point because it's yet another barrier to people trying to submit PRs (on top of keeping docs up to date).
I generally try to make up for this by trying my best not to create breaking changes.  Can you give some examples of problems you ran into lately?

AllanWang:
@Schmavery I still haven't updated my bot yet but I know we were talking about threadHistory having arguments that don't work, leading to [this](https://github.com/Schmavery/facebook-chat-api/commit/ad25b6b60a252c1b8c3f39780cfe369bde7bb495) PR.

If you're okay with it, I can create a Changelog.md and maintain it at least for the summer (and likely during the school year as well). If Github releases are created again I can link to the specific changes/new features so it will be easier for others to update.

Schmavery:
@AllanWang yeah that's a good example of the type of change that comes along now and then where the old version doesn't work anymore because of a change on facebook's end, which forces us to change the signature.

If you want to start a simple Changelog.md that covers the last couple releases that would be great, and I'll try to encourage people to add to the top of it whenever we make any breaking changes.  I'll have to see if I can make some time to go back and add the github release tags where I missed them

AllanWang:
@Schmavery could you make a release tag now just so I can link things to a specific commit when I make the log? Otherwise I'll link it to the latest commit upon creation rather than to a release

Schmavery:
@AllanWang Just tagged a couple, let me know if I missed something.

ravkr:
@Schmavery this should be closed now

PR #435, opened 2017-03-30
Attachment forwarding

GAMELASTER:
You can forward various attachments for multiple users.
Actually I tested videos and images. GIFs should work too.
Can anyone test forwading files and audio files?

PR #481, opened 2017-05-09
Changelog

AllanWang:
Initial changelog support
The changelog is meant to highlight changes that matter to the users.
Changes involve breaking updates, implementation changes, etc. Internal updates are not listed unless they affect how a user might write their code.


AllanWang:
Addresses #480

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 ## 1.4.0 (04-28-2017)
-* [Breaking change] [getThreadHistory](/DOCS.md#getThreadHistory): update parameters
-* [New] [setMessageReaction](/DOCS.md#setMessageReaction)
+* [Breaking change] [getThreadHistory](/DOCS.md#getThreadHistory): update parameters - no more start & end params; replaced with amount
```
ravkr: Maybe "use amount and timestamp now"?

AllanWang: My reason for that was because start & end was replaced by amount. The [commit](https://github.com/Schmavery/facebook-chat-api/commit/ad25b6b60a252c1b8c3f39780cfe369bde7bb495) shows that timestamp was already part of the function, so it's not a new parameter

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
...
 ## 1.4.0 (04-28-2017)
 * [Breaking change] [getThreadHistory](/DOCS.md#getThreadHistory): update parameters - no more start & end params; replaced with amount
-* [New] [setMessageReaction](/DOCS.md#setMessageReaction) - set reaction on message
-* [New] [forwardAttachment](/DOCS.md#forwardAttachment) - send attachment to array of users
-* [setMessage](/DOCS.md#apisendmessagemessage-threadid-callback): added Mentions field
+* [New] [setMessageReaction](/DOCS.md#setMessageReaction): set reaction on message
+* [New] [forwardAttachment](/DOCS.md#forwardAttachment): send attachment to array of users
+* [setMessage](/DOCS.md#sendMessage): added Mentions field
 * [sendTypingIndicator](/DOCS.md#sendTypingIndicator): make callback optional
+* [getUserID](/DOCS.md#getUserID): support searches for more than just names, and add a type parameter
```
ravkr: I would write "now returns all results instead of just users" because it was throwing error before
And as far as I remember "type" is not a parameter but a field in result object (?)

AllanWang: Oops you're right. I'll change that

ravkr:
How about keeping dates in YYYY-MM-DD instead of MM-DD-YYYY? I think it is more popular in IT because of [ISO_8601](https://en.wikipedia.org/wiki/ISO_8601)

AllanWang:
@ravkr  done

CHANGELOG.md
ravkr: I know that it is not new, but it would make it clear to other users if they haven't used timestamp before (only start and end) 
@Schmavery what do you think?

Schmavery:
Maybe it's too much to ask, but it would be great if, when possible, these link to the relevant PRs so that people can go figure out exactly what changed and get some more info.

AllanWang:
@Schmavery I don't mind at all. I had it like that originally but wasn't sure if it would be better to link to the newest release. I'll do that im a bit

Schmavery:
Awesome, I'm imagining just a little "#13123123" at the end of the change description when applicable.

gamelaster:
I thinks about writing into changelog which bugfix fixing for example "Listening in NaN" bug

ravkr:
This a good idea

AllanWang:
I'll have a look at that. Let me know if I miss anything later

bobland1:
Are we sticking with American date or shall it be DDMMYY instead

AllanWang:
@gamelaster do you have any particular commits I should look at? Not sure which ones you are referring to

ravkr:
AFAIK you can write `[#453]`instead of `[#435](https://github.com/Schmavery/facebook-chat-api/pull/435))`

AllanWang:
The only issue is if someone forks it and others look at the changelog it will show the request on their repo (which doesn't exist) rather than on the original

Schmavery:
Great, thanks a lot @AllanWang

ravkr:
maybe I was wrong.. :/ links are not working
zrzut ekranu z 2017-05-10 22 48 29

AllanWang:
@Schmavery can we revert the [latest commit](https://github.com/Schmavery/facebook-chat-api/pull/481/commits/feef4bc4d2e34eeb3a81ca2e8bce79249a426e52)? I thought the links would work on your repo but it seems like it doesn't.

Schmavery:
Hmm, I see.  Can you submit a quick PR with the links changed back when you get a chance?  I squashed the PR and I'm not sure how to do the git gymnastics necessary to revert just part of it at this point.

ravkr:
@AllanWang how about you removing latest commit and @Schmavery rebasing repository? (its not an old commit... :/ )

AllanWang:
see #482 
I rebased again and added the commit

Schmavery:
Thanks :)

Issue #396, opened 2017-02-11
searchForThread could not find thread name


lyneca:
The [docs](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#apisearchforthreadname-callback) for `searchForThread` say that it can be used to search for "chat titles", but when I search for the name of one of my group chats, it returns `Could not find thread [name]`. Is the documentation old or am I just using it wrong?

ravkr:
You're right, I just tried to search by thread name and nothing is found... but when I'm putting e.g. name of one of the participants then I get array of threads with that user :/

lyneca:
Must be a Facebook thing then - this has worked in the past. I'll poke around in the chrome dev tools and see what I can find

lyneca:
Ok it has changed, Facebook now sends a call to messenger.com/api/graphqlbatch/ which returns three JSON strings - one for users that match your search, one for threads that match your search, and a paging info string.

It's a get request with form data - the important bits of data that I could see are these:
```
'batch_name': 'MessengerTypeaheadResultRoute',
'queries':
'{
  "q3": {
    "priority": 0,
    "q": "Query MessengerTypeaheadResultRoute {entities_named([my search]) {id,@F4}} QueryFragment F0 : User {id,name,is_memorialized,is_viewer_friend,is_viewer_coworker,profile_picture.width(32).height(32) as _profile_picture4QyiK {uri},messenger_contact {is_on_viewer_contact_list,id},url,is_messenger_user,is_work_user,viewer_affinity} QueryFragment F1 : Node {id,__typename} QueryFragment F2 : Page {id,name,profile_picture.width(32).height(32) as _profile_picture4QyiK {uri},url,is_messenger_user} QueryFragment F3 : Application {id,name,instant_game_info {icon_uri}} QueryFragment F4 : SearchableEntitiesQuery {search_results.of_type(user).include_messenger_only_results_for_web_messenger(true).first(20) as _search_resultsLr77o {edges {node {__typename,@F0,@F1},cursor},page_info {has_next_page,has_previous_page}},search_results.messenger_commerce(true).include_games(true).first(5) as _search_resultsSJD9D {edges {node {__typename,@F2,@F3,@F1},cursor},page_info {has_next_page,has_previous_page}},id}",
    "query_params": {}
  },
  "q4": {
    "priority": 0,
[...]

PR #488, opened 2017-06-12
Update thread fields in utils.formatThread

### LINKED ISSUE #485: getThreadInfo formerParticipants is null
After removing a user from a group, formerParticipants for that group is still null. How is this property updated and does Facebook even give this information? I tried looking into this myself but I wouldn't know how to test it and got confused since getThreadInfo uses getUserInfo but getUserInfo's info does have the same properties.

Thanks!

mangopearapples:
This PR changes some of the properties in `formatThread`. I removed some deprecated fields and added the missing fields that were available in the payload but not to the user.

This also somewhat resolves #485 by removing the field from the DOCS and `formatThread`

Notable changes:

- `muteSettings` -> `muteUntil`: is the timestamp for which the thread is muted until. -1 is forever.
- -`canonicalFbid`
- +`lastMessageTimestamp` + `lastReadTimestamp`
- +`adminIDs`
- +`threadType`


Schmavery:
Thanks for updating this, @mangopearapples 
As always when we're updating the public interface like this, is it possible to get any more info about what is in these new fields?  It would be amazing if we could update the docs to have this information too, but that might be out of scope for this change (I would give lots of emoji hearts to whoever takes on that task).
If facebook has stopped serving some of this data, I'm ok to remove it, but we should try to be at least a little careful not to add more data people won't know how to use?

mangopearapples:
You're right about adding data people won't know how to use. Should I remove them? I had somewhat considered this and ignored some properties like `ephemeral_ttl_mode` but I though fields like adminIDs and `lastMessageTimestamp` were self explanatory. 

Are there any particular fields you should not be added?

I will update docs to include description.

Schmavery:
isPinProtected/adminIDs/threadType/reactionsMuteMode/mentionsMuteMode
Maybe these correspond to thread features I'm unaware of?

Schmavery:
Also, searching through the repo, it looks like formatThread is used by 3 functions: getThreadList, getThreadInfo, and searchForThread.  Which of these were you able to test with the changes?  Do we receive all the values in all these cases?

mangopearapples:
@Schmavery Are doc formats like [this](https://github.com/mangopearapples/facebook-chat-api/blob/hotfix/formatThread-fields-fix/DOCS.md#apigetthreadinfothreadid-callback) okay?

Also from what's available to me, adminIDs are members that are admins in groups in chats belonging to a Facebook group. E.g. .

`threadType` seems to be '1' for normal user chats and '2' for group chats and maybe different values for chats from pages. More testing is needed for this but I think it is useful data for the user.

I was assuming mentions and reactions can be muted in the settings but maybe not. I will remove isPinProtected, reactionsMuteMode, and mentionsMuteMode. Is this okay?

I have tested with `getThreadList` and `getThreadInfo` which give identical values (apart from `getThreadList` being an array of course). I was unaware of `searchForThread` so more testing on that needs to be done.

mangopearapples:
After a bit of research, doesn't look like Facebook uses `search_threads.php` anymore. Instead, it uses GraphQL. Issue #396 seems to correspond with this too. I can't get `searchForThread` to give me any thread data so I can't know if a successful call will give me appropriate values.

Schmavery:
Great, thanks a lot for following up on this.  Threadtype looks very interesting, we were using getUserInfo in a hacky way to get this info, but maybe we can switch over now.  I wonder how much this overlaps with isCanonical (I never knew what a canonical thread was until now).
Thanks for trying out getThreadList and getThreadInfo.  The doc changes are great.

DOCS.md
```diff
@@ -516,9 +516,43 @@ Takes a threadID and a callback.  Works for both single-user and group threads.
...
+| unreadCount | Number of unread messages |
+| messageCount | Number of messages |
+| imageSrc | URL to the group chat photo. Null if unset or a canonical thread. |
+| timestamp |  |
+| serverTimestamp |  |
+| muteUntil | Timestamp at which the thread will no longer be muted. The timestamp will be -1 if the thread is muted indefinitely or null if the thread is not muted. |
+| mentionsMuteMode |  |
+| isCanonicalUser | True if the other user in a canonical thread is a user, false if the other user is a page or group. |
+| isCanonical | True if the thread is a private chat to a single user, false for group chats. |
+| isPinProtected |  |
```
Schmavery: Should this be removed now

mangopearapples: Yep forgot about that one. Done

Schmavery:
Thanks!

Issue #484, opened 2017-05-21
Facebook-Chat-api not talking with Messenger Platform Bot


miketamis:
When i try to get the facebook-chat-api to send a message to a messenger platform bot, I get the following error
```
ERR! getUserInfo { __ar: 1,
ERR! getUserInfo   __sf: 'iw',
ERR! getUserInfo   error: 1357004,
ERR! getUserInfo   errorSummary: 'Sorry, something went wrong',
ERR! getUserInfo   errorDescription: 'Please try closing and re-opening your browser window.',
ERR! getUserInfo   payload: null,
ERR! getUserInfo   bootloadable: {},
ERR! getUserInfo   ixData: {},
ERR! getUserInfo   lid: '6422424341438859408' }
```

Schmavery:
We have no tested support for this, as far as I know.  What is the usecase?

miketamis:
The use case is to test a messenger bot in production to get response time stats

hoangdevelopers:
I also have the same error when bot reply messenger.  i guess phantom had problem. but i dont know clear phantom cache
![image](https://cloud.githubusercontent.com/assets/12141921/26302635/156caaec-3f0f-11e7-9ce8-b5694694eb73.png)

Schmavery:
@hoangdevelopers this api doesn't use phantomjs, I'm not sure what you're talking about.  This error is very generic.  Are you also trying to communicate with a messenger platform bot?

@miketamis I think it wouldn't be a bad thing to support this, but I don't currently understand how these messenger platform bots are implemented, so we need to do some investigation.  Can you open your browser dev tools and inspect the network calls that are made when sending a message to a messenger platform bot?  If you could share those with us, it might help.

codeforequity-at:
Some input on this topic: my test automation framework TestMyBot (https://github.com/codeforequity-at/testmybot) supports running automated conversations against a messenger platform bot. I'm actually using this library for the nuts&bolts.
I built a sample project: https://github.com/codeforequity-at/testmybot/tree/master/samples/fbdirect

I didn't do anything special, just a plain sample Facebook Messenger Bot deployed on Heroku.

Schmavery:
@miketamis any chance this is working for you now too?

bsansouci:
Moving discussion to #500

PR #493, opened 2017-06-18
Use require() than JSON.parse(fs.readFileSync())

JacobGunther12:


bsansouci:
Hey @JacobGunther12 Thanks for the PR!
But I don't think this is a very important change. You seem eager to help though. If so, could you look at https://github.com/Schmavery/facebook-chat-api/issues/484 and test whether you can talk to a bot or not? That would really help us and maybe we can close that issue.

Thanks :)

Schmavery:
Another good option if you don't know where to start with Ben's suggestion might be https://github.com/Schmavery/facebook-chat-api/issues/463.  If you have any questions you can definitely ask on that issue and it will help everyone 😄 
I think for this PR, I wrote it this way in order to be more explicit about the fact that that file is JSON and not full JS, fwiw.

PR #503, opened 2017-07-18
Add MessageAudio attachment type

### LINKED ISSUE #383: Fix support for voice message attachments
All in the title, is it possible to get a recorded voice message (in stream ? .waw ?)

mangopearapples:
Adds `MessageAudio` attachment type to GraphQL version of `getThreadHistory` (`getThreadHistoryGraphQL`). I've tried to keep and copy the naming conventions of the fields from the other attachment types

Think this fixes #383 as well.

I've discard some fields for the user and kept what I think is appropriate but I'd be happy to add/remove any fields that you might think should be changed.

### DIFF
--- a/src/getThreadHistoryGraphQL.js
+++ b/src/getThreadHistoryGraphQL.js
@@ -97,6 +97,18 @@ function formatAttachmentsGraphQLResponse(attachment) {
         contentType: attachment.content_type,
         filename: attachment.filename,
       }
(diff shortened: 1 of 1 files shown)


Schmavery:
We're probably going to be playing with all the signatures a little here anyway so I'll merge this to save you the trouble of rebasing.  Thanks!

PR #510, opened 2017-08-09
This will allow people to easily get mentioned users in a message obj

cthuluhoop123:


Schmavery:
Hey @cthuluhoop123, thanks for contributing. I've got a few questions I'm hoping you can help me with.
Why does it default to empty string rather than empty array? Is `m.delta.data.prng` always defined when `m.delta.data` is defined?
Is there currently no way to see mentioned users in the message object? What is the format of the mentions this PR returns? An array of userids?
Can you include an update to the docs and changelist?

cthuluhoop123:
@Schmavery Sorry. The empty string is a mistake, I didn't catch that. Currently the api does not provide a way to easily grab the (unfortunately, only the ids) users mentioned in a message. This array returns the mentions in the order that they are in the message. I'm on mobile now so I'll make a few changes in a few hours after school. I'll add all this to the docs and changelist

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

cthuluhoop123:
``m.delta.data`` does not always contain a mention property (such as if the message simply doesn't mention anyone), hence the tertiary operator.

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Schmavery:
I see a conditional check to see if `m.delta` contains a `data` property, but not to see if that `data` contains a `prng`.

cthuluhoop123:
From my testing, ``m.delta.data`` will always contain prng. However I'll add a check just in case.

cthuluhoop123:
Also, what ways are there to avoid being blocked on facebook?

bsansouci:
Thanks!

For the way to avoid being blocked:
- don't send too many messages too quickly
- don't send too many messsages to strangers
- watch out for where the account's IP address is, FB will detect sudden changes and will make you verify that you are you

cthuluhoop123:
Thanks man :) pleasure working with you.

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Issue #506, opened 2017-08-01
Changing thread colour not working? 


bobland1:
Anyone else being not able to use the change colour feature? I don't reseave and error but nor does the bot change the thread colour?

Theblood:
I have the same problem @bobland1 :'(

bobland1:
Yeah it's confusing, I'll do some testing. See what is actually happening as at the moment it doesn't give either response (error / work)

Schmavery:
Thanks for looking into it, let us know if you need help or find anything out.

bobland1:
Nothing shows up even with verbose, not sure what's going on to be honest

AstroCB:
I have been running into this issue as well for the past month or so. It appears only to be a problem for hex values that are not in the palette on the actual website/app UI, so Facebook may have altered their backend to prevent the usage of custom values outside of those provided by them.

If this is the case, it would be nice to be able to programmatically access these acceptable values in some way (whether as a string with the name of the color or a list of hex values).

Schmavery:
Hmm, good to know @AstroCB .  Can you share one or two values that you've found work so that bobland1 can see if this is the issue?

AstroCB:
Apologies, it appears that those values no longer work either.

If you'd like to test them yourself, you can try these colors (which are all in the palette): `#44bec7`, `#fa3c4c`, and `#d4a88c`.

Schmavery:
I don't see any changes in my browser environment that would indicate this being broken.  Can someone experiencing this problem check what message is getting sent in the browser to change color?

bobland1:
Mauf - Custom Messenger Colors was the way you were able todo it. That is now not working and the creator is working on a fix.

bobland1:
Request URL:https://www.messenger.com/messaging/save_thread_color/?&dpr=1&source=thread_settings
Request Method:POST
Status Code:200 
Remote Address:157.240.1.18:443
Referrer Policy:origin-when-cross-origin
Response Headers
access-control-allow-credentials:true
access-control-allow-method:OPTIONS
access-control-allow-origin:https://www.messenger.com
access-control-expose-headers:X-FB-Debug, X-Loader-Length
cache-control:private, no-cache, no-store, must-revalidate
content-encoding:br
[...]

PR #512, opened 2017-08-10
Add getValidThreadColors

AstroCB:
Adds a function to get accepted values for `changeThreadColor` and validation within `changeThreadColor` to prevent undefined behavior for non-supported colors as per #506.


Schmavery:
Thanks for contributing some more changes :)
Is there any reason you're not just exposing the data directly? As in:
```
    return ["#44bec7", "#ffc300", "#fa3c4c", "#d696bb", "#6699cc", ...];
};
```

I'm also wondering if it wouldn't be more readable if we actually made it a map of the human-readable names exposed by facebook to the colors. The python clone of this api [does this](https://github.com/carpedm20/fbchat/blob/master/fbchat/models.py#L185). i.e.
```
{
    MessengerBlue = ''
    Viking = '#44bec7'
    GoldenPoppy = '#ffc300'
    RadicalRed = '#fa3c4c'
    Shocking = '#d696bb
    ...
}
```
Might end up with more readable code and make it more possible for us to add/adapt/remove colors as time goes on (though I'll admit the names are... strange at best).

AstroCB:
Ah, I haven't seen any constant (non-functional) data exposed via the API so I wasn't sure if that would be against the design you guys have set up.

Are those names the official ones exposed by Facebook? I haven't seen them anywhere.

AstroCB:
I switched it to a dictionary using those names; the code should definitely be more readable that way.

bsansouci:
Seems good to me though I'd change the docs to say that it is an object rather than it returns an object.

DOCS.md
```diff
@@ -715,6 +716,29 @@ login({appState: JSON.parse(fs.readFileSync('appstate.json', 'utf8'))}, (err, ap
 
 ---------------------------------------
 
+<a name="validThreadColors"></a>
+### api.validThreadColors
+
+Returns a dictionary mapping names of all currently valid thread colors to their hexadecimal values that are accepted by [`api.changeThreadColor`](#changeThreadColor). These colors, listed below, are the ones present in the palette UI used for selecting thread colors on the Messenger client. Due to Facebook backend changes, the thread color can no longer be set to an arbitrary hex value.
```
Schmavery: Maybe we can move the note about "Due to Facebook backend changes, the thread color can no longer be set to an arbitrary hex value." to the CHANGELOG.md file and put a link there to this PR?  (See the other entries for an example)

Schmavery:
I feel like the validity is implied by the fact that we're providing the colors, so just `threadColors` is probably more concise.  Thanks!

AstroCB:
Done.

Schmavery:
Seems good, thanks a lot.

PR #521, opened 2017-08-23
Use a default name for unavailable accounts

### LINKED ISSUE #520: getThreadHistory fails when one message author is no longer available
To reproduce:

1. use page B to talk with user A
2. unpublish page B
3. call getThreadHistory on this thread from user A

will fail at this line https://github.com/Schmavery/facebook-chat-api/blob/master/src/getThreadHistory.js#L51

as `api.getUserInfo` can't get info of page B

xuezhma:
Fixes #520

Use 'Facebook User' as sender_name for no longer available accounts


Schmavery:
Thanks for this quick fix for your issue!

PR #526, opened 2017-09-10
Multiple message callbacks for listen (close #525)

### LINKED ISSUE #525: Multiple message callbacks for listen
Hey !
When we use the method `listen`, we can use only one callback.  

To explain, i have this example:
```javascript
api.setOptions({
        listenEvents: true,
        logLevel: "silent"
});
// After...
api.listen((err, message) => {
    console.log("not called");
});
api.listen((err, message) => {
    console.log("called");
});
```  
and the output is:
```  
info login Logging in...
info login Logged in
info login Request to reconnect
info login Request to pull 1
info login Request to pull 2
info login Request to thread_sync
info login Done logging in.
called
called
called
```

I think that's a good idea to add multiple 'listener' of message :)

Vinetos:
Now we can call several times `listen` to use more than one callbacks
Also close #525

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -8,8 +8,10 @@ var identity = function() {};
 
 module.exports = function(defaultFuncs, api, ctx) {
   var currentlyRunning = null;
-  var globalCallback = identity;
+  var globalCallback = [];
+  globalCallback.push(identity);
 
+  // Stop all listening
   var stopListening = function() {
     globalCallback = identity;
     if(currentlyRunning) {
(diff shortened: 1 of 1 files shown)


QuenK1:
Is very good idea ! :D

Schmavery:
Hi @Vinetos, thanks for contributing! As I said in #525, I'm not convinced this is necessary -- this functionality can be very easily accomplished in user code.
Is there a use case that I'm missing that is impossible/hard to accomplish without this change?

Schmavery:
See also my example in https://github.com/Schmavery/facebook-chat-api/issues/525#issuecomment-329016053

Schmavery:
Closing as explained in #525, feel free to comment if you have any issues.

Issue #524, opened 2017-09-07
No longer able to use API: Cookie not in this host's domain


alecmanabat:
ERR! login Error: Cookie not in this host's domain. Cookie:www.facebook.com Request:www.messenger.com

Any idea what is causing this/how to fix?

Schmavery:
If I remember right, we use messenger.com for some calls and facebook.com for other calls. Maybe we need to be more careful about what cookies we're sending. Seems strange though. Do you see this during a specific call?
cc @bsansouci

alecmanabat:
I'm not sure when it occurs exactly, but seems to be right when the boot tries to log in. Can't do testing until later, but for now I attached a part of the console log that captures the error.
[max-chatbot-logs-1504799617701.txt](https://github.com/Schmavery/facebook-chat-api/files/1285686/max-chatbot-logs-1504799617701.txt)

Schmavery:
Hmm ok, if this happens before login completes, that helps narrow it down a little. Thanks

alecmanabat:
Thanks for looking into it, bot was running fine until last night. I can post my code too if the log isn't enough

RandallKent:
I also started seeing this error on September 7th. 

A call to `getFriendsList()` results in:

```
ERR! login Error: Cookie not in this host's domain. Cookie:www.facebook.com Request:www.messenger.com
ERR! login     at CookieJar.setCookie (/app/node_modules/facebook-chat-api/node_modules/request/node_modules/tough-cookie/lib/cookie.js:941:13)
ERR! login     at CookieJar.setCookieSync (/app/node_modules/facebook-chat-api/node_modules/request/node_modules/tough-cookie/lib/cookie.js:1305:18)
ERR! login     at RequestJar.setCookie (/app/node_modules/facebook-chat-api/node_modules/request/lib/cookies.js:26:20)
ERR! login     at /app/node_modules/facebook-chat-api/utils.js:725:11
ERR! login     at Array.forEach (native)
ERR! login     at /app/node_modules/facebook-chat-api/utils.js:720:13
ERR! login     at tryCatcher (/app/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:26:23)
ERR! login     at Promise._settlePromiseFromHandler (/app/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:507:31)
[...]

PR #536, opened 2017-10-13
Fix #531

### LINKED ISSUE #531: Error Not logged in
Hi, 

I have an issue since yesterday. Indeed, when I want to set a reaction to an message, create a poll or just get user info, I have this error in my logs : 

`ERR! setReaction { error: 'Not logged in.' }`

`ERR! getUserInfo Error: parseAndCheckLogin got status code: 404. Bailing out of trying to parse response.`

But I am connected yet : I can receive and send messages.

The problem is maybe caused by the api ? 

Thank you

knyzorg:
This code isn't even written by me. The frenchman in #531 had the fix for it but it has been weeks and it wasn't merged or PR'd. Hence, I am opening this PR for it to be fixed at last.

It doesn't break anything as far as I see.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -783,6 +783,8 @@ function saveCookies(jar) {
       if (c.indexOf(".facebook.com") > -1) {
         jar.setCookie(c, "https://www.facebook.com");
       }
+      var c2 = c.replace(/domain=\.facebook\.com/, "domain=.messenger.com");
+      jar.setCookie(c2, "https://www.messenger.com");
     });
     return res;
   };


Schmavery:
The reason this hasn't been merged or PRed is that it breaks something else and we unfortunately haven't found the right fix that solves both issues. Definitely open to suggestions here but I think flip-flopping between having these lines and not having them is a bad approach here. Thanks for helping out!

https://github.com/Schmavery/facebook-chat-api/issues/524
As you can see, removing those exact two lines was the fix for this issue.

knyzorg:
What functions does this break, exactly? As far as I see, this project aims to support Messenger, not facebook. Facebook already has APIs they offer.

The `getFriendsList()` shouldn't be given more importance over the reactions as reactions are a component of messenger and friends are not.

Schmavery:
I agree that we should prioritize messenger support over facebook support.  Issue #524 reported having trouble *logging in* because of this issue, which is generally even higher priority. Other than that, I have no more information than is stated in the issue, as I was unable to reproduce the problem.

For some background, this api was initially written to work with calls against facebook.com.  As messenger.com became more used and gained features not supported by facebook.com, we attempted to add support for them.  Many of the cookies were in common so simply renaming them was an easy hack that got us 90% of the way there.  Clearly this is being done incorrectly in some way, so that should be investigated. I have trouble believing that reintroducing this hack is the right choice.

knyzorg:
Were you able to reproduce the login issue? I haven't.

I occasionally get login issues because facebook throttles me if I log in/out too often. Maybe that was the issue?

Schmavery:
It's unfortunately the nature of this work that it's frequently not possible to reproduce real issues. We are forced to rely heavily on the users of the api to explain their problems in detail. 
@alecmanabat very helpfully included his logs of the failure (which many people neglect to do), in which I see
```
2017-09-07T15:53:24.906668+00:00 app[web.1]: starting bot!
2017-09-07T15:53:30.914568+00:00 app[web.1]: ERR! login Error: Cookie not in this host's domain. Cookie:www.facebook.com Request:www.messenger.com
```

Login throttling issues are common as you say, but this doesn't look like any login throttling issue I've ever seen.  That combined with it going away when those lines were removed seemed to be fairly telling.
Anyone looking at these two lines of code can see that they were clearly a hack. They were introduce in the hope that they would solve a problem, which they seem to have done, at least in part, but it seems to me that they have outlived their usefulness and are due to be replaced with code that actually attempts to retrieve cookies from messenger.com

alecmanabat:
Sorry if this is off topic. I probably should have posted this in #524 while it was being discussed.
I uncommented the two lines of code in question, and removed all other code from my bot except: `login({
    appState: JSON.parse(fs.readFileSync(config.stateFile, "utf8"))
}, (err, api) => {
    if (err) return console.error(err);

    api.setOptions({
        logLevel: "silent",
        forceLogin: true,
        listenEvents: false
    });
    console.log("login complete");
});`
I still got the same error in the logs I posted before. The "login complete" message does not print. I was hoping that it was just an issue with getFriendsList(), since I used that right after the bot logged in and I was mistaken thinking it was a login issue, but it appears that is not the case.

knyzorg:
Does the issue persist if you remove `forceLogin` and `appState` in favor of normal login? Just curious.

alecmanabat:
The appState file was causing the problem, generating a new one without using any code besides login fixed it. I don't know too much about how the appState file works, but maybe after saving the cookies with the "hack" causes problems with future logins.

knyzorg:
@Schmavery 

If the above is the case, then the login routine for when appstate is used should be rewritten. Ideally, enforce normal auth and use appstate and normal auth as a fallback.

Schmavery:
This is some great investigation, thanks @alecmanabat @knyzorg 
Can someone inspect the appstate file with and without the cookie hack and see if there's any substantial difference?

alecmanabat:
I attempted investigating the appstate file. When I commented 5 days ago I could reproduce the login error when logging in with the appstate file even with new appstate files generated after a credentials login. Now I cannot reproduce the error with the cookie hack still in place and calling `getFriendsList`. Maybe something on Facebook's end fixed this? I haven't been following #531 but perhaps the error from #524 is no longer a problem. `forceLogin` does not seem to make a diifference either.

Schmavery:
Thanks for trying again @alecmanabat.  Interesting that you can no longer repro.
@RandallKent, you were also running into this error. Any chance you can try again as well before we merge these lines back in?

Schmavery:
Since we don't seem to have a repro of this issue, I'm going to re-merge it with fingers crossed

PR #546, opened 2017-10-29
Fixed the parsing of facebook attachments now that they have changed …

### LINKED ISSUE #544: Attachments need to be updated to new format
I'm getting the following error when an image is sent to the API user:

```
ERR! listen Error: unrecognized attach_file `{"blob_attachment":{"__typename":"MessageImage","attribution_app":null,"attribution_metadata":null,"filename":"image-1962166334031351","preview":{"uri":"<URL to the image>","height":280,"width":497},"large_preview":{"uri":"<URL to the image","height":480,"width":853},"thumbnail":{"uri":"<URL to the image>"},"legacy_attachment_id":"1962166334031351","original_dimensions":{"x":1920,"y":1080},"original_extension":"png","render_as_sticker":false}}`
ERR! listen     at _formatAttachment (<PATH>node_modules/facebook-chat-api/utils.js:354:13)
ERR! listen     at attachments.map.v <PATH>node_modules/facebook-chat-api/utils.js:376:55)
ERR! listen     at Array.map (<anonymous>)
ERR! listen     at Object.formatDeltaMessage (/home/sarithis/nodejs/own/chat666/node_modules/facebook-chat-api/utils.js:376:46)
ERR! listen     at resolveAttachmentUrl (<PATH>node_modules/facebook-chat-api/src/listen.js:159:40)
ERR! listen     at resolveAttachmentUrl (<PATH>node_modules/facebook-chat-api/src/listen.js:168:30)
ERR! listen     at parsePackets (<PATH>node_modules/facebook-chat-api/src/listen.js:171:19)
ERR! listen     at Array.forEach (<anonymous>)
ERR! listen     at <PATH>node_modules/facebook-chat-api/src/listen.js:109:12
ERR! listen     at tryCatcher (<PATH>node_modules/bluebird/js/main/util.js:26:23)
ERR! listen     at Promise._settlePromiseFromHandler (<PATH>node_modules/bluebird/js/main/promise.js:510:31)
ERR! listen     at Promise._settlePromiseAt (<PATH>node_modules/bluebird/js/main/promise.js:584:18)
ERR! listen     at Promise._settlePromises (<PATH>node_modules/bluebird/js/main/promise.js:700:14)
ERR! listen     at Async._drainQueue (<PATH>node_modules/bluebird/js/main/async.js:123:16)
ERR! listen     at Async._drainQueues (<PATH>node_modules/bluebird/js/main/async.js:133:10)
ERR! listen     at Immediate.Async.drainQueues (<PATH>node_modules/bluebird/js/main/async.js:15:14)
```

It's fixable by adding the following case statement in the _formatAttachment function in utils.js:
```
case undefined:
  return {
    type: "error",

    // Save error attachments because we're unsure of their format,
    // and whether there are cases they contain something useful for debugging.
    attachment1: attachment1,
    attachment2: attachment2
};
```
The attachment1 object doesn't contain an attach_type property, that's why it's undefined.

mitchcapper:
…it, closes #544 
did everything but video sends as I do not have a device that does that.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -236,10 +236,19 @@ function _formatAttachment(attachment1, attachment2) {
   // Instead of having a bunch of if statements guarding every access to image_data,
   // we set it to empty object and use the fact that it'll return undefined.
   attachment2 = attachment2 || {id:"", image_data: {}};
-  var fileName = attachment1.filename
-  attachment1 = attachment1.mercury ? attachment1.mercury : attachment1
-
-  switch (attachment1.attach_type) {
+  var fileName = attachment1.filename;
+  attachment1 = attachment1.mercury ? attachment1.mercury : attachment1;
+  var blob = attachment1.blob_attachment;
+  var type =  blob && blob.__typename ? blob.__typename : attachment1.attach_type;
+  if (! type && attachment1.sticker_attachment){
+    type = "StickerAttachment";
+    blob = attachment1.sticker_attachment;
+  }
+  else if (! type && attachment1.extensible_attachment){
+    type = "ExtensibleAttachment";
+    blob = attachment1.extensible_attachment;
+  }
+  switch (type) {
     case "sticker":
       return {
         type: "sticker",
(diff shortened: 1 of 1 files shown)


utils.js
```diff
@@ -236,10 +236,21 @@ function _formatAttachment(attachment1, attachment2) {
...
-  var fileName = attachment1.filename
-  attachment1 = attachment1.mercury ? attachment1.mercury : attachment1
-
-  switch (attachment1.attach_type) {
+  var fileName = attachment1.filename;
+  attachment1 = attachment1.mercury ? attachment1.mercury : attachment1;
+  var blob = attachment1.blob_attachment;
+  var type = blob ? blob.__typename :  attachment1.attach_type;
+  if (! type)
+    type = attachment1.attach_type;
```
Schmavery: It seems like this and the `var type = blob ? blob.__typename :  attachment1.attach_type;` above are redundant. Am I missing something?

Schmavery:
This looks awesome! Thanks for submitting a fix so quickly. Made one comment but after that should be good to merge 😄

utils.js
Schmavery: Ah, I guess it handles the case that blob exists but blob.__typename is null/undefined. Could do it in one line with `blob && blob.__typename ? blob.__typename :  attachment1.attach_type`, but I guess I'll leave that call to you.

mitchcapper:
Your solution was better, updated PR.

utils.js
```diff
@@ -350,8 +359,91 @@ function _formatAttachment(attachment1, attachment2) {
...
+        subattachments: blob.story_attachment.subattachments,
+        width: blob.story_attachment.media.image.width,
+        height: blob.story_attachment.media.image.height,
+        image: blob.story_attachment.media.image.uri,
+        playable: blob.story_attachment.media.is_playable,
+        duration: blob.story_attachment.media.playable_duration_in_ms,
+        source: blob.story_attachment.source.text,
+        title: blob.story_attachment.title_with_entities.text,
+        facebookUrl: blob.story_attachment.url,
+        target: blob.story_attachment.target.__typename,
```
bsansouci: Are you sure this should be `__typename` here?

mitchcapper: Well generate some data and see if you think something else.  I had limited content sharing options to try for but it seemed the field that matched it best.   Either add additional debug lines or rename the case to something else and it will auto-dump the entire data stream when you share something.  If you can provide the json that also works:)

bsansouci: Ok I just checked. When you share any link you'll get
```
"target": {
  "__typename": "ExternalUrl"
}
```

This indicates that there is such a thing as an "internal" URL, given that it's FB I tried sharing a URL to an FB event:
```
    "target": {
      "__typename": "Event",
      "eventID": "116707585671845",
      "guestStatus": null,
      "canViewerJoin": true,
      "timeString": "Friday, November 10 at 5:00pm",
      "event_place": {
        "name": "Dogpatch Boulders"
      },
      "social_context": {
        "text": "Fitness · 666 people"
      },
      "watchStatus": "UNWATCHED",
      "connection_style": "INTERESTED",
      "canViewerWatch": true
    },
```

I couldn't find other objects that had different fields, so let's do something like:
```
target: blob.story_attachment.target == null ? undefined : {
  type: blob.story_attachment.target.__typename,
  eventID: blob.story_attachment.target.eventID,
  guestStatus: blob.story_attachment.target.guestStatus,
  canViewerJoin: blob.story_attachment.target.canViewerJoin,
  timeString: blob.story_attachment.target.timeString,
  eventPlaceName: blob.story_attachment.target.event_place == null ? null : blob.story_attachment.target.event_place.name,
  socialContext: blob.story_attachment.target.social_context == null ? null : blob.story_attachment.target.social_context.text,
  watchStatus: blob.story_attachment.target.watchStatus,
  connectionStyle: blob.story_attachment.target.connection_style,
  canViewerWatch: blob.story_attachment.target.canViewerWatch
}
```

mitchcapper: Target before was not a complex structure (I don't believe) as such I would not suggest making it a complex structure.  I agree on providing more details however, so maybe add a new field with target_details with that structure?  What do you think.  Mainly I want to try and break as few as applications as possible that used the data before when updating the lib?

Schmavery: As far as I remember, the target before was simply an unprocessed object from facebook and the closest imitation would probably be to simply use `blob.story_attachment.target`. Afaict there's no evidence that `target` was ever a string.

However, I think what @bsansouci is getting at is that it would be nice to start processing this object so that it can be more consistent. I'm not sure that this necessarily needs to be part of this fix, but you can definitely feel free to.
If we do this, I think we would need a switch on `blob.story_attachment.target.__typename` and do something like what @bsansouci is suggesting in `case "Event":`, with some exception or warning logging in the case where we don't recognize the `__typename`.
If we do this, then a section would probably need to be added under the *Attachments* section [here](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#apilistencallback) for types of share. Technically the fields of the attachments should be documented a little more in the future but I won't make you do that 😛

mitchcapper: Ah, well then I have no problem making the object his object if people want that in the PR.

It would be good if someone setup a FB account with a bunch of test posts (or links to posts?) part of what I ran into an issue with testing the interface was finding all the content types to generate the structures.  As I mention in the PR, I didn't even see how to send video through the web interface to update that structure.

Schmavery: In order to avoid too much back and forth on this PR (which can already help people in its current form 🎉 ), I'd love it if we could just change target to be `blob.story_attachment.target` and then merge this, followed up with a second PR for any additional fixes, if you have time/willingness :)

mitchcapper:
ok I believe this is what people were asking for:) If someone gets me the json for a video send I can do the target structure and video at the same time for a diff PR.

bsansouci:
Ok let's do what @Schmavery suggested. 
Thanks a lot :D <3

PR #548, opened 2017-10-31
Login approvals fix

### LINKED ISSUE #139: Login approvals are broken
I'm having trouble getting the login process to work.

I have second factor and login approvals turned on. When attempting to login with my regular password using login(), the first callback has the following error:

```
Couldn't login. Facebook might have blocked this account. Please login with a browser or enable the option 'forceLogin' and try again.
```

My Facebook account is not blocked. I do not get a prompt to input my second factor. I tried to set `forceLogin` to `true` but it didn't help.

forcemagic:
> **Update**: It's fixed now! (somehow..)

It's almost done, but I still have a problem.

The thing is, when I try to log in, it says `Login successful`, but whenever I make calls to the API, it says... Okay, let this picture speak for itself:

error desc

For the record, I approved from another device, that's why I put nothing in the approvals code field. If you have the authenticator or SMS setup, you can certainly use them.

Fixes #139


Schmavery:
Awesome, this already looks like a great step in the right direction, thanks.
I'm not sure what could be causing the 404 but I'll try to think about it some more today.
Interesting that you can approve from another device, didn't know that.  Do you just submit nothing as the code in that case?

Side note: do you mind removing the package-lock.json? Probably accidentally got included in the commit :)

forcemagic:
Yes, you can approve from another device, and yes, if that's the case, you should put "" (nothing) into the 2FA code field.

By the way, does this 404 error happen on a normal account?

forcemagic:
Oh, and the last 2 calls (or so) about verifying the device turned out to be unnecessary.

forcemagic:
Miraculously, it works now. I have no idea how, but... yeah. I'll remove the `WIP` tag. :smile:

Schmavery:
That's great! It looks like a very safe change so I'll merge it and if there continue to be problems we can do more exploration.

bobland1:
Really welldone @forcemagic. Long time waiting

PR #550, opened 2017-11-03
Add support for MessageVideo and MessageAudio attachments

ivkos:
Related to #544, #546

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -398,6 +398,30 @@ function _formatAttachment(attachment1, attachment2) {
         animatedWebpUrl: blob.animated_image.uri,
         animatedWebpPreviewUrl: blob.preview_image.uri,
       };
+    case "MessageVideo":
+      return {
+        type: "video",
+        filename: blob.filename,
+        thumbnailUrl: blob.large_image.uri,
+        previewUrl: blob.large_image.uri,
+        previewWidth: blob.large_image.width,
+        previewHeight: blob.large_image.height,
+        ID: blob.legacy_attachment_id,
+        url: blob.playable_url,
+        width: blob.original_dimensions.x,
+        height: blob.original_dimensions.y,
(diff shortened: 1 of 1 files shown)


utils.js
```diff
@@ -398,6 +398,30 @@ function _formatAttachment(attachment1, attachment2) {
...
+        previewHeight: blob.large_image.height,
+        ID: blob.legacy_attachment_id,
+        url: blob.playable_url,
+        width: blob.original_dimensions.x,
+        height: blob.original_dimensions.y,
+        duration: blob.playable_duration_in_ms,
+      };
+    case "MessageAudio":
+      return {
+        type: "audio",
```
Schmavery: Can you add a quick entry to the attachments section of https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#listen to add an "audio" type?

utils.js
```diff
@@ -398,6 +398,30 @@ function _formatAttachment(attachment1, attachment2) {
...
+        audioType: blob.audio_type,
+        isVoiceMail: blob.is_voicemail,
+        filename: blob.filename,
+        ID: blob.url_shimhash,
+        url: blob.playable_url,
+        durationInMs: blob.playable_duration_in_ms,
```
Schmavery: Can we change this to just "duration" to match the other attachment types?

Schmavery:
This looks awesome, thanks for fixing. Will merge as soon as comments are addressed.

utils.js
Schmavery: Ah, I see you were matching the existing audio type that's not documented...
Any change you can tweak that one to also use duration rather than durationInMs?
Clearly I didn't read that PR close enough.  If you don't have time, I can just merge this and fix it myself.

ivkos:
Can you please merge it and fix them, because I'm not near a PC. Thanks!

PR #552, opened 2017-11-05
Add error handling for ExtensibleAttachment

laszlolm:
I was testing the plugin with facebook business messages (best way to generate messages for yourself is to create a page and enter messenger's tutorial where you message the page)

Unfortunately bot media types are missing from the plugin and while it's like that I added some extra error handling (on cards the app was completely crashing or in fact, any rich message (with buttons etc))

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -442,20 +442,20 @@ function _formatAttachment(attachment1, attachment2) {
     case "ExtensibleAttachment":
       return {
         type: "share",
-        description: blob.story_attachment.description.text,
-        ID: blob.legacy_attachment_id,
-        subattachments: blob.story_attachment.subattachments,
-        width: blob.story_attachment.media.image.width,
-        height: blob.story_attachment.media.image.height,
-        image: blob.story_attachment.media.image.uri,
-        playable: blob.story_attachment.media.is_playable,
-        duration: blob.story_attachment.media.playable_duration_in_ms,
-        source: blob.story_attachment.source.text,
-        title: blob.story_attachment.title_with_entities.text,
-        facebookUrl: blob.story_attachment.url,
-        target: blob.story_attachment.target,
-        styleList: blob.story_attachment.style_list,
-        url: blob.story_attachment.url
+        description: blob.story_attachment.description ? blob.story_attachment.description.text : '',
+        ID: blob.legacy_attachment_id ? blob.legacy_attachment_id : 0,
+        subattachments: blob.story_attachment.subattachments ? blob.story_attachment.subattachments : '',
+        width: (blob.story_attachment.media && blob.story_attachment.media.image) ? blob.story_attachment.media.image.width : 0,
+        height: (blob.story_attachment.media && blob.story_attachment.media.image) ? blob.story_attachment.media.image.height : 0,
+        image: (blob.story_attachment.media && blob.story_attachment.media.image) ? blob.story_attachment.media.image.uri : '',
+        playable: (blob.story_attachment.media && blob.story_attachment.media.is_playable) ? blob.story_attachment.media.is_playable : false,
+        duration: (blob.story_attachment.media && blob.story_attachment.media.playable_duration_in_ms) ? blob.story_attachment.media.playable_duration_in_ms : 0,
+        source: blob.story_attachment.source ? blob.story_attachment.source.text : '',
+        title: blob.story_attachment.title_with_entities ? blob.story_attachment.title_with_entities.text : '',
+        facebookUrl: blob.story_attachment.url ? blob.story_attachment.url : '',
+        target: blob.story_attachment.target ? blob.story_attachment.target : '',
+        styleList: blob.story_attachment.style_list ? blob.story_attachment.style_list : null,
+        url: blob.story_attachment.url ? blob.story_attachment.url : ''
       };
     case "MessageFile":
       return {


Schmavery:
What does the attachment that you're having trouble formatting look like? Can you show an example?

laszlolm:
@Schmavery its mainly messenger platform messages. If you create a new page on Facebook and messages it as the owner you get this tutorial. Also every bot works with these elements. Look at the attached pictures:
screen shot 2017-11-06 at 08 08 04
screen shot 2017-11-06 at 08 09 31

bsansouci:
Can you share the json?

laszlolm:
```
{ legacy_attachment_id: '1131669340300774',
  story_attachment:
   { description: null,
     media: null,
     source: { text: 'Publisher app' },
     style_list: [ 'h_scroll', 'fallback' ],
     title_with_entities:
      { text: 'Forms, Content types, Human input understanding, Broadcast messages, Import content' },
     properties: [],
     url: null,
     deduplication_key: 'ffd215bf73dfff9735e897f48e6fa788',
     action_links: [],
     messaging_attribution: null,
     messenger_call_to_actions: [],
     target:
      { __typename: 'MessengerBusinessMessage',
        message: 'Expect: Forms',
        call_to_actions: [],
        business_items: [Object] },
     subattachments: [ [Object], [Object], [Object], [Object], [Object] ] } }
{ type: 'message',
  senderID: '1099240420210333',
  body: '',
  threadID: '1099240420210333',
  messageID: 'mid.$cAAO9M8WWpedlwXjuJ1fkDOj5O7J6',
  attachments:
   [ { type: 'share',
       description: '',
       ID: '1131669340300774',
       subattachments: [Array],
       width: 0,
       height: 0,
       image: '',
       playable: false,
       duration: 0,
       source: 'Publisher app',
       title: 'Forms, Content types, Human input understanding, Broadcast messages, Import content',
       facebookUrl: '',
       target: [Object],
       styleList: [Array],
       url: '' } ],
  mentions: {},
  timestamp: '1509952824871',
  isGroup: false } null
{ name: 'Expect', id: '1099240420210333' }
```

laszlolm:
This is the return value of `blob` variable in the main menu

laszlolm:
This picture (forms, content types...):
image

mitchcapper:
I can confirm my initial implementation does not work with other share formats.  I tried to find some others and the PR here does seem to handle them better.  We are going to either end up with a lot of these tests I feel or a safe accessor like discussed in #553 to handle the various fb share options.

laszlolm:
All of these special formats revert to ExtensibleAttachment, after a week of using it there were no errors I found

Schmavery:
Sorry all, been pretty busy lately so haven't had a chance to take a closer look at this PR.

I'm always wary of making blanket changes like this because the tradeoff is like this:

Without error handling (as it's defined by this PR):
 - Users have to handle the occasional erroring object (which they already do by checking for `err` in the callback)

With error handling:
- We need to make sure we have reasonable defaults for all values, which can be hard (for example, IDs). If we fail here, then users of the api have to manually do checks of the value of each field before using it (i.e. check whether `msg.image === ''` before using it).
- We (as users and a community trying to maintain this api) become blind to changes on facebook's end. You actually *want* the message parse to fail if the signature changes because then you'll be able to notice that the endpoint has changed and you're not getting the data you think you're getting. We can definitely make nothing error and silently return objects full of `''`, but it will be hard to use that to build a bot/tool etc.

That said, we do do some of this sort of error handling in various location in the api, and it's a bit of a fine line to try and walk when designing. Sometimes objects returned by facebook have different fields in different scenarios, and we do lots of work to unify these based on their semantics when possible. For example, we have a number of places in the code like this:
```
threadID: formatID((originalMessage.thread_fbid || originalMessage.other_user_fbid).toString()),
```

Whenever possible, we want to identify which fields can change or be absent in an object we are expecting, and then substitute them with another meaningful field of the object whenever possible, and failing that, try to assign a sensible default. If no sensible default exists, generally it's likely that the two types of message object we're trying to fit into one response should actually be exposed to the user as two separate types of object as they're semantically different enough. Otherwise.. maybe an error actually needs to be thrown in that case, haha

I want to make it clear that everyone really appreciates the effort that was put in here to make the changes and fix the api. This message is not really directed at the writer of this change but also to several other comments I've seen about this type of issue lately. I think there was some good effort put in here to come up with good defaults for values and I just want to make sure we're all on the same page design-wise. I get concerned by the fact that this entire type of object could "parse successfully" and come back to the user completely empty. Maybe there's some subset of this change that will address the problem you're experiencing without risking making the object potentially useless.

mitchcapper:
I get the idea of we want things to break loudly when the API changes so we know to fix it.  At the same token there is certainly something of value for continuity of the app.

I think 90%+ of people would prefer things were just degraded than fail completely on API change.

I think better would be a boolean setting, or just an error call for invalid messages.   Yes obviously just handling a message that can't be parsed to an error function is fine, but better would be to be able to return a partial message.

Granted returning a partial message (or detecting this) means more code.  We could simply have the safe parser have a bool of if a default was used and if so then flag the message as partially processed.

Just some thoughts but I can tell you that a minor API change on FB requiring me to sit down and code that instant to get the chat API working is not my idea of fun:)   Even if it is better for the library for it to get fixed fast. Some hairy warnings might be enough.

Schmavery:
> "just an error call for invalid messages"

I'm pretty sure we do that? Is https://github.com/Schmavery/facebook-chat-api/commit/7552ce69561ac424cf6461e4c0cd8762ae2a7b9f not working? If it's not, we should definitely fix it 😄

mitchcapper:
Sorry I was saying a separate function/event for partially parsed messages, so a user knows that messages coming that way have issues.

bsansouci:
@mitchcapper We pass you the original data that we received from FB in the error, so you can take the fields you really want from it. But I'd recommend against doing that because if the API is failing it means we are not officially supporting something.
While partial messages would allow you to not handle errors gracefully, it'll most likely make your app silently worse on every breaking change FB does.

The best would be to pin point those problems as they appear so we can keep the API up and running.
@laszlolm would you mind going through the fields you're receive that break your app and adding checks for those, so we stop the crash. I'll merge that!

mitchcapper:
@bsansouci of course, but re-implementing parsing logic would be redundant.   I agree partially parsed messages is not good for any long term solution but if the api separated out distribution of them it would be clear to users they may be broken.  Imagine the FB api changes the dimensions spec for images to another field.   All the sudden all image attachments would break, even though the image URL may be the same and that may be all the application needs/cares about.   

As this API parses more and more metadata that increases the likelihood of one of the fields will change and decreases the chance that change will be relevant to the app.    This is not a public API so FB has no desire or need to avoid changing it at will.

Another flip side is we should have a "basic" parsing of messages and then an extended parse option that gets other fields.    That way apps could degrade semi-gracefully rather than losing complete parsing due to a minor field change.

bsansouci:
@mitchcapper I understand your perspective. I would argue this would achieve nothing because the end results would be the same. Say the URL field of images changes, and now it's empty string. How long until you notice? Let's imagine you notice immediately, how do you fix it? You need to have access to the underlying object and do some "parsing" yourself such that your app continue to behave correctly. You can do exactly this today, by looking at the error we return you and doing any patching you want accordingly. In either case you'll want to do a PR to this API so you don't have to maintain custom parsing functions.

> As this API parses more and more metadata that increases the likelihood of one of the fields will change and decreases the chance that change will be relevant to the app. This is not a public API so FB has no desire or need to avoid changing it at will.

We are very aware of that. We are stepping towards our third year maintaining this library without major breakages in the last year and half. 

If there are any fields that you're missing or data causing a crash for your app I would more than happy to fix it, or merge a PR that fixes that. Closing this issue unless something new comes up.

PR #554, opened 2017-11-07
listen: Better handle errors thrown from formatDeltaMessage

ivkos:
This is a workaround for the issue that caused the console to get
infinitely spammed with the same error over and over again.

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -156,7 +156,12 @@ module.exports = function(defaultFuncs, api, ctx) {
               if (v.delta.class == "NewMessage") {
                 (function resolveAttachmentUrl(i) {
                   if (i == v.delta.attachments.length) {
-                    var fmtMsg = utils.formatDeltaMessage(v);
+                    var fmtMsg;
+                    try {
+                      fmtMsg = utils.formatDeltaMessage(v);
+                    } catch (err) {
+                      return globalCallback(err);
+                    }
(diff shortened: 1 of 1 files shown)


mitchcapper:
@ivkos  thanks.  It would be good if we returned a more complex object to identify the issue (parseerror), include the original message we were trying to parse, and the original exception rather than just the exception.  That would allow for better control by implementers to handle error types.

Schmavery:
@mitchcapper something like this?
```
return globalCallback({
    error: "Problem parsing attachment type. Please open an issue at https://github.com/Schmavery/facebook-chat-api/issues.",
    ex: err,
    data: v});
```

mitchcapper:
To match the structure of other error calls I would rename the fields as:
ex->detail,
data->res

I would also suggest we add another field error_type = "ATTACHMENT_PARSE_ERROR" or similar.  This does not follow other errors but would make it easier for developers looking for this error (rather than matching the long string).

mitchcapper:
Sorry, did not realize who posted the last comment, use his structure rather than my suggestion.

Schmavery:
Always glad to consider more opinions, let's go with @mitchcapper's solution.
```
return globalCallback({
    error: "Problem parsing attachment type. Please open an issue at https://github.com/Schmavery/facebook-chat-api/issues.",
    detail: err,
    res: v
    type: "ATTACHMENT_PARSE_ERROR"});
```

Issue #534, opened 2017-09-29
getThreadHistory no longer working?


Beaudan:
Some time in the last couple days it seems that getThreadHistory has stopped working as intended. If supplied no timestamp and asked for 10 messages, the first 10 messages of a thread are returned. If supplied with a timestamp however, the first 11 messages of the thread are returned no matter what the timestamp is. Very recently this was working in my application to retrieve the number of messages requested starting from the timestamp supplied and going backwards.

Is this happening for anyone else or am I somehow doing something really stupid?

kieferlam:
If it has been working before, it's most likely Facebook changing their API. Probably to Graph.

Try `getThreadHistoryGraphQL`

Schmavery:
@Beaudan and let us know how it works! Then we can swap over to using graphql by default.

Medexstar:
Hi, I work with @Beaudan on our application using this api. Tried out the getThreadHistoryGraphQL, and it works almost perfectly. I'm not exactly sure how all the query stuff works, but using the same api arguments (threadID, amount, timestamp, callback), it gave similar results to the original getThreadHistory api endpoint. However there were a couple of message/event type cases that were missed and thus ended in error, which I added, and now our application works perfectly again.

I might make a PR soon with my changes to accommodate for missing message/event types.

However the event data is different to the original api result (just some minor differences, mostly naming, such as no logMessageData/Body/Type, and instead it returns something similar to a regular message result). See here:
https://github.com/Schmavery/facebook-chat-api/blob/520dc53dd404422ae6a6c1a3c4b73d28d5883dde/src/getThreadHistoryGraphQL.js#L277-L292
for an example, where "ParticipantsAddedMessage" is similar to the "log:subscribe" event type in the original api. For now we just had to change our frontend application with the naming/data changes.

ravkr:
getThreadHistory uses GraphQL now

@Schmavery, we can close it too

PR #556, opened 2017-11-12
Add support for many type of events and MessageAnimatedImage attachments

KevinSalmon:
I tried to export 200 000 messages from my messenger conversations. getThreadHistory does not work currently (#534) so I used the new version getThreadHistoryGraphQL(). I have some parsing errors which I have fixes (mainly null fields or not detected).


bsansouci:
Sorry for having taken so long to merge this. Thanks for the contribution :)

PR #562, opened 2017-12-09
getThreadInfoGraphQL.js:formatExtensibleAttachment() fixes

jrwiebe:
* Made several attachment field values conditionally null if specified element does not exist.

Previously, `formatExtensibleAttachment()` was failing on certain types of attachments -- i.e., attachments with missing `attachment.story_attachment.source` or `attachment.story_attachment.media` fields.


Leo4815162342:
why not merge?

Schmavery:
I think for some reason I thought this was adding new fields but if it's just fixes then I don't see any reason why not.  Sorry about the delay guys and thanks for the reminder!

PR #571, opened 2018-01-10
Fix parsing of facebook story attachments

ivankolesnikov:
The description and media fields can be null.


Schmavery:
Is an extensibleattachment useful without an attachment? Unfortunately we seem to have made the assumption that it wasn't. We can probably accept this change @bsansouci

ivankolesnikov:
@Schmavery According to https://developers.facebook.com/docs/messenger-platform/reference/template/generic#elements the image_url field is optional and in some rare cases we rely on this behavior.

bsansouci:
Cool thanks for this!

PR #572, opened 2018-01-12
Add getThreadList using GraphQL

ravkr:
as old getThreadList stopped working (and now returns only empty array) just started making new version

there is separate field with id for group conversations and "one to one" conversations...
should I merge them to just one "threadID"?

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -639,52 +639,187 @@ __`participants` format__
 - `"ReducedMessagingActor"`
 (*there might be more*)
 
-| Key                       | Description                                                   |
-|---------------------------|---------------------------------------------------------------|
-| accountType               | `"User"`                                                      |
-| userID                    | ID of user                                                    |
-| name                      | Full name of user                                             |
-| shortName                 | Short name of user (most likely first name)                   |
-| gender                    | Either `"MALE"`, `"FEMALE"` or `"NEUTER"`                     |
-| url                       | URL of the user's Facebook profile                            |
-| profilePicture            | URL of the profile picture                                    |
-| username                  | Username of user or `null`                                    |
-| isViewerFriend            | Is the user a friend of you?                                  |
-| isMessengerUser           | Does the user use Messenger?                                  |
-| isVerified                | Is the user verified? (Little blue tick mark)                 |
-| isMessageBlockedByViewer  | Is the user blocking messages from you?                       |
-| isViewerCoworker          | Is the user your coworker? *(not yet tested)*                 |
-| isEmployee                | `null`? *(not yet tested)*                                    |
+<table>
+<tr>
+<th>Account type</th>
+<th>Key</th>
+<th>Description</th>
+</tr>
+<tr>
+<td rowspan="13">User</td>
+<td>userID</td>
+<td>ID of user</td>
+</tr>
+<tr>
+<td>name</td>
+<td>Full name of user</td>
+</tr>
+<tr>
+<td>shortName</td>
+<td>Short name of user (most likely first name)</td>
+</tr>
+<tr>
+<td>gender</td>
+<td>Either         
+<code>"MALE"</code>,         
+<code>"FEMALE"</code> or         
+<code>"NEUTER"</code>
+</td>
+</tr>
+<tr>
+<td>url</td>
+<td>URL of the user's Facebook profile</td>
+</tr>
+<tr>
+<td>profilePicture</td>
+<td>URL of the profile picture</td>
+</tr>
+<tr>
+<td>username</td>
+<td>Username of user or         
+<code>null</code>
+</td>
+</tr>
+<tr>
+<td>isViewerFriend</td>
+<td>Is the user a friend of you?</td>
+</tr>
+<tr>
+<td>isMessengerUser</td>
+<td>Does the user use Messenger?</td>
+</tr>
+<tr>
+<td>isVerified</td>
+<td>Is the user verified? (Little blue tick mark)</td>
+</tr>
+<tr>
+<td>isMessageBlockedByViewer</td>
+<td>Is the user blocking messages from you?</td>
+</tr>
+<tr>
+<td>isViewerCoworker</td>
+<td>Is the user your coworker?         
+<em>(not yet tested)</em>
+</td>
+</tr>
+<tr>
+<td>isEmployee</td>
+<td>
+<code>null</code>?         
+<em>(not yet tested)</em>
+</td>
+</tr>
 
+<tr>
+<td rowspan="10">Page</td>
+<td>userID</td>
+<td>ID of the page</td>
+</tr>
+<tr>
+<td>name</td>
+<td>Name of the fanpage</td>
+</tr>
+<tr>
+<td>url</td>
+<td>URL of the fanpage</td>
+</tr>
+<tr>
+<td>profilePicture</td>
+<td>URL of the profile picture</td>
+</tr>
+<tr>
+<td>username</td>
+<td>Username of user or         
+<code>null</code>
+</td>
+</tr>
+<tr>
+<td>acceptsMessengerUserFeedback</td>
+<td></td>
+</tr>
+<tr>
+<td>isMessengerUser</td>
+<td>Does the fanpage use Messenger?</td>
+</tr>
+<tr>
+<td>isVerified</td>
+<td>Is the fanpage verified? (Little blue tick mark)</td>
+</tr>
+<tr>
+<td>isMessengerPlatformBot</td>
+<td>Is the fanpage a bot</td>
+</tr>
+<tr>
+<td>isMessageBlockedByViewer</td>
+<td>Is the fanpage blocking messages from you?</td>
+</tr>
 
-| Key                       | Description                                                   |
-|---------------------------|---------------------------------------------------------------|
-| accountType               | `"Page"`                                                      |
-| userID                    | ID of the page                                                |
-| name                      | Name of the fanpage                                           |
-| url                       | URL of the fanpage                                            |
-| profilePicture            | URL of the profile picture                                    |
-| username                  | Username of user or `null`                                    |
-| acceptsMessengerUserFeedback |                                                            |
-| isMessengerUser           | Does the fanpage use Messenger?                               |
-| isVerified                | Is the fanpage verified? (Little blue tick mark)              |
-| isMessengerPlatformBot    | Is the fanpage a bot                                          |
-| isMessageBlockedByViewer  | Is the fanpage blocking messages from you?                    |
-
-
-| Key                       | Description                                                   |
-|---------------------------|---------------------------------------------------------------|
-| accountType               | **1** `"ReducedMessagingActor"` (account requres verification, messages are hidden) or **2** `"UnavailableMessagingActor"` (account disabled/removed) |
-| userID                    | ID of the user                                                |
-| name                      | **1** Name of the user or **2** *Facebook User* in user's language |
-| url                       | `null`                                                        |
-| profilePicture            | URL of the default Facebook profile picture                   |
-| username                  | Username of user or `null`                                    |
-| acceptsMessengerUserFeedback |                                                            |
-| isMessageBlockedByViewer  | Is the user blocking messages from you?                       |
+<tr>
+<td rowspan="7">ReducedMessagingActor<br />(account requres verification,<br />messages are hidden)</td>
+<td>userID</td>
+<td>ID of the user</td>
+</tr>
+<tr>
+<td>name</td>
+<td>Name of the user</td>
+</tr>
+<tr>
+<td>url</td>
+<td>
+<code>null</code>
+</td>
+</tr>
+<tr>
+<td>profilePicture</td>
+<td>URL of the default Facebook profile picture</td>
+</tr>
+<tr>
+<td>username</td>
+<td>Username of user</td>
+</td>
+</tr>
+<tr>
+<td>acceptsMessengerUserFeedback</td>
+<td></td>
+</tr>
+<tr>
+<td>isMessageBlockedByViewer</td>
+<td>Is the user blocking messages from you?</td>
+</tr>
+<tr>
+<td rowspan="7">UnavailableMessagingActor<br />(account disabled/removed)</td>
+<td>userID</td>
+<td>ID of the user</td>
+</tr>
+<tr>
+<td>name</td>
+<td><em>Facebook User</em> in user's language</td>
+</tr>
+<tr>
+<td>url</td>
+<td><code>null</code></td>
+</tr>
+<tr>
+<td>profilePicture</td>
+<td>URL of the default Facebook profile picture</td>
+</tr>
+<tr>
+<td>username</td>
+<td><code>null</code></td>
+</tr>
+<tr>
+<td>acceptsMessengerUserFeedback</td>
+<td></td>
+</tr>
+<tr>
+<td>isMessageBlockedByViewer</td>
+<td>Is the user blocking messages from you?</td>
+</tr>
+</table>
 
 
 In a case that some account type is not supported, we return just this *but you can't rely on it* and log warning to the console
+
 | Key          | Description             |
 |--------------|-------------------------|
 | accountType  | type, can be anything   |
--- a/DOCS.md
+++ b/DOCS.md
@@ -579,49 +579,118 @@ __Arguments__
 ---------------------------------------
 
 <a name="getThreadList"></a>
-### api.getThreadList(start, end, type, callback)
+### api.getThreadList(limit, timestamp, tags, callback)
 
-Will return information about threads.
+Returns information about the user's threads.
 
 __Arguments__
 
-* `start`: Start index in the list of recently used threads.
-* `end`: End index.
-* `type`: Optional String, can be 'inbox', 'pending', or 'archived'. Inbox is default.
-* `callback(err, arr)`: A callback called when the query is done (either with an error or with an confirmation object). `arr` is an array of thread object containing the following properties:
+* `limit`: Limit the number of threads to fetch.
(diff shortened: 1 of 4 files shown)


ravkr:
how about name of thread? When it is not set Facebook says it is `null`, should I replace it with an empty string? and what to do in case of  1 to 1 conversations? Put name of another person? an empty string? `null`?

ravkr:
there is `"outgoing_bubble_color": "FF8C0077"` should we return `FF8C0077`, `8C0077` or `#8C0077`?

Schmavery:
Thanks for working on this @ravkr 
The best guidance I can likely give is to look at the format* functions in utils.js and try to be consistent with those where possible.
formatMessage is probably one of the most stable, there we have these answers to your questions:
```js
threadID: formatID((originalMessage.thread_fbid || originalMessage.other_user_fbid).toString()),
threadName: (originalMessage.group_thread_info ? originalMessage.group_thread_info.name : originalMessage.sender_name),
```

(Of course, formatThread doesn't seem to follow the same convention but I think that's not by design and that the behavior in formatMessage is generally preferable)

As for the color, I think we have classically returned whatever facebook gives us, but I could see an argument for leaving out the initial `FF` (which is the alpha value, I guess)... not sure.

Schmavery:
@ravkr are you still working on this PR or is it waiting for more review/merging? It would be really great if we could replace the broken getThreadList with a (fully?) compatible call made to graphql.

ravkr:
I didn't have much time this week, sorry... but I will try to finish it on Sunday or Monday

forcemagic:
You need some help with this @ravkr? I'm happy to assist if you have any problems :smiley:

ravkr:
@forcemagic you can try to write documentation to that function
here is example how to use it https://pastebin.com/NAfZYiU5 - it will just dump current formatted data to a file (but it is not complete yet)

ravkr:
can someone inspect what values can be in `rtc_call_data`?
i got only `NO_ONGOING_CALL` and `AUDIO_GROUP_CALL` but when it is not a group call I am also getting `NO_ONGOING_CALL`

_some examples from Facebook responses_

    "rtc_call_data": {
        "call_state": "NO_ONGOING_CALL",
        "server_info_data": "",
        "initiator": {
            "id": "100000000000000"
        }
    },

    "rtc_call_data": {
        "call_state": "NO_ONGOING_CALL",
        "server_info_data": "",
        "initiator": null
    },

    "rtc_call_data": {
        "call_state": "AUDIO_GROUP_CALL",
        "server_info_data": "GANsbGEYFkdST1VQOjIwMDAwMDAwMDAwMDAwMDAYCFlxZnhHU25nAA==",
        // decoded: "\u0018\u0003lla\u0018\u0016GROUP:2000000000000000\u0018\u0008YqfxGSng\u0000"
        "initiator": {
            "id": "100000000000000"
        }
    },

Schmavery:
@ravkr the existing getThreadList doesn't have that field so it's fine to leave it out (unless you have a reason to really need it?)

ravkr:
Thank you @forcemagic 👍 I've just merged it
@Schmavery I know that original getThreadList didn't have it, but why ignore something that Facebook send us? this new version returns few things more, but we can omit unknown data until we have a clue what it does 

there is one bigger change... now the functions returns `participants` - array with all users from conversation, along with info about them (this way calling `getUserInfo` for every userID is now obsolete)
of course function name is different... but should I map `participants` to make old `participantIDs` usable as before?

forcemagic:
Happy to help @ravkr :) if I see any new changes I'll try to keep the docs up to date to reflect them.

ravkr:
there is not much left to do....
obraz
(these are the currently ignored things)

theweavrs:
@ravkr Is this completely implemented?

Schmavery:
@ravkr what are the differences between what this call returns and what the original getThreadList returns? I would much rather prefer to replace the broken call than add a new one with a note saying the old one is broken

ravkr:
@Schmavery these were removed by facebook: `serverTimestamp`, `isCanonical`, `isCanonicalUser`, `recipientsLoadable`, `hasEmailParticipant`, `readOnly`, `canReply`, `lastMessageType`
and now `participants` contains array of participants along with data about them (as Facebook sends it anyway), this is the reason for removing `participantIDs`... (you can still get IDs from `participants`)

any other fields should be compatible with the old version (correct me if I'm wrong)

Schmavery:
If those are the only differences and we add a participantIDs field that is derived from participants, we can probably replace the broken function with this.
@bsansouci do you agree?

forcemagic:
We could make an "alias" for every GraphQL query and make them backwards compatible. (Example: `getThreadList()` would actually call `getThreadListGraphQL()`. @Schmavery if you're okay with this, I would like to update the docs (and a bit of code), so that it reflects this approach.

ravkr:
IMO `getThreadList()` should call `getThreadListGraphQL()` and print some warning message

Schmavery:
Why do we need both?

ravkr:
well... I'd put both for 2-3 months so people will get answer right in their console. otherwise some of them would probably create issues on github `why did you remove getThreadList?!?!? are there any alternatives!??!`

Schmavery:
Except that as far as we can tell, getThreadList doesn't work. Anyway, we wouldn't remove getThreadList, we'd remove the test version suffixed with GraphQL.

forcemagic:
@Schmavery We could extend `getThreadList()` with the new fields and somehow extract values for the old fields, so we don't need a new function, but we have already started creating `*GraphQL()` funcs, so... Who knows?

Schmavery:
The plan was always to rename the existing *GraphQL functions (you'll notice they are not documented). Unfortunately they were written as prototypes/tests of using graphql so they need a little bit of compatibility work before they can be switched over. I'm just trying to save us that work now :)

ritu99:
Does getThreadList() with graphQL need any more work? I can try and help out if needed.

ravkr:
@ritu99 it is almost done but I do not have much time :( 
you can try to add support for different folders (look for `"tags":["INBOX"]` in javascript) and/or do some tests if everything works in "weird configurations" (talks with bots, banned users, disabled accounts, accounts waiting for verification, threads with people who blocked you etc.)

ALiangLiang:
I discover two event types named "VideoCallMessage" and "VoiceCallMessage".

ravkr:
from several messenger requests I've extracted these `tags`:
```
["ARCHIVED"]
["INBOX"]
["INBOX","unread"]
["OTHER","unread"]
["PENDING","unread"]
```

are there any other folders?

ravkr:
I've just found two new account types `"UnavailableMessagingActor"` and `"ReducedMessagingActor"`

Schmavery:
When this is ready, please rename it from getThreadListGraphQL to getThreadList and update the docs accordingly :)

ravkr:
@Schmavery should I change `threadType: t.thread_type` to `isGroup:` with `true` or `false`?

somebody could also review the code, it should be almost complete now

Schmavery:
As we did in with the other graphQL functions, please add `isGroup: bool` and remove `threadType` from the docs but keep `threadType` on the actual object (with the same semantics) marked with `// @Legacy` so we can remove it in the next major version update.  Will try to do more review when I get a chance.

ravkr:
how about now? is everything okey?

forcemagic:
For me, everything seems fine, let's see what @Schmavery says :smile:

Budgetperson:
Quick comment as this as I was trying to use it early :). I think https://github.com/Schmavery/facebook-chat-api/pull/572/files#diff-017a6e261d15146e15fe740821e4aa47R104 needs to be changed to `isGroup: t.thread_type === "GROUP"`. The current implementation is causing an error for me. Thanks for the work on this PR!

ravkr:
@Budgetperson thanks, fixed now

ravkr:
support for `adminIDs` added

DOCS.md
```diff
@@ -579,49 +579,255 @@ __Arguments__
...
+<td>Is the user verified? (Little blue tick mark)</td>
+</tr>
+<tr>
+<td>isMessageBlockedByViewer</td>
+<td>Is the user blocking messages from you?</td>
+</tr>
+<tr>
+<td>isViewerCoworker</td>
+<td>Is the user your coworker?
+<em>(not yet tested)</em>
```
Schmavery: For these "not yet tested" ones, can we just leave them out of the docs if we don't know what they do?

Bjornskjald: It's tested, it returns `true` if user is in the same company on workplace (see [here](https://github.com/ravkr/facebook-chat-api/files/1849633/workplace.json.txt))

ravkr:
@Schmavery what do you think?

Bjornskjald:
@Schmavery can you merge it? I'd like to use new getThreadList, but I'm not 100% sure about stability of @ravkr's repo (if it won't be deleted after merge)...

Schmavery:
@ravkr thanks for working on this.

Let's not document untested fields (isEmployee, montageThread). We can add these later if someone understands and needs them.

I did a very quick comparison with the old getThreadList and see that we no longer have participantIDs. Can we add this and other missing fields to maintain compatibility? The same would probably apply between the format of the new participants field and the return value of getUserInfo, though I haven't had a chance to compare those myself yet.

There are several comments like `or maybe not? not tested` in the actual implementation. Can we get the signature of this function sorted out in general (and match the signature of the previous call, where we can)? Some things: make sure tags/type is optional, handle the case where it's a string, check the passed in type/tags values to make sure the user doesn't pass in something invalid.

Some of the logic around handling timestamp is not clear to me.

`esData[0].o0.data.viewer.message_threads.nodes` looks pretty brittle, are we sure this object will remain the same?

Bjornskjald:
@Schmavery 
> see that we no longer have participantIDs

> now the functions returns participants - array with all users from conversation, along with info about them (this way calling getUserInfo for every userID is now obsolete)
of course function name is different... but should I map participants to make old participantIDs usable as before?
https://github.com/Schmavery/facebook-chat-api/pull/572#issuecomment-363220305

Schmavery:
@Bjornskjald https://github.com/Schmavery/facebook-chat-api/pull/572#issuecomment-364579242

Bjornskjald:
Uh, okay, haven't seen it...

ravkr:
@Schmavery 
`participantIDs` should not be mentioned in the docs? Am I right?

these `or maybe not? not tested` are there because Facebook uses it, but we never can be sure if something else would work... :/  But I've just removed these comments

`resData[0].o0.data.viewer.message_threads.nodes` should work all the times
`resData[0]` is just the response (`[1]` is object with info about error/successful results)
`.o0.data` is result of our request (notice that we are sending `"o0"` in our `form` object)
etc..
obraz
It doesn't look like it would change anytime soon

I've explained logic around `timestamp` in the comments in code
```js
// When we ask for threads using timestamp from the previous request,
// we are getting the last thread repeated as the first thread in this response.
// .shift() gets rid of it
// It is also the reason for increasing limit by 1 when timestamp is set
// this way user asks for 10 threads, we are asking for 11,
// but after removing the duplicated one, it is again 10
```

ravkr:
@Schmavery is everything okey now?

delmotte:
when will this be available ? empty array is really boring...

ravkr:
@Schmavery

lgnap:
@Schmavery ?

bsansouci:
My fault! I told him to not merge until I look at it. So looking at it today I promise ;)

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+function formatColor(color) {
+  if (color && color.match(/^(?:[0-9a-fA-F]{8})$/g)) {
+    return color.slice(2);
+  }
+  return color;
+}
 
-    if (end <= start) end = start + 20;
+function getThreadName(t) {
+  if (t.name || t.thread_key.thread_fbid) return t.name;
```
bsansouci: This looks weird, why do we return `t.name` even when it's null/undefined?

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
-    };
+  for (let p of t.all_participants.nodes) {
+    if (p.messaging_actor.id === t.thread_key.other_user_id) return p.messaging_actor.name;
+  }
+}
 
-    form[type + "[offset]"] = start;
-    form[type + "[limit]"] = end - start;
+function mapNicknames(customizationInfo) {
+  return customizationInfo?customizationInfo.participant_customizations.map(u => {
```
bsansouci: can you do `customizationInfo && customizationInfo.participant_customizations ? ...`

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+    return {"userID": u.participant_id, "nickname": u.nickname};
+  }):[];
+}
 
-    if (ctx.globalOptions.pageID) {
-      form.request_user_id = ctx.globalOptions.pageID;
+function formatThreadList(data) {
+  return data.map(t => {
+    return {
+      threadID: utils.formatID(t.thread_key.thread_fbid || t.thread_key.other_user_id),
```
bsansouci: can you do `t.thread_key ? (t.thread_key.thread_fbid || t.thread_key.other_user_id) : null`

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+      // isCanonicalUser: t.is_canonical_neo_user, // is it always false?
+      // TODO: how about putting snippet in another object? current implementation does not handle every possibile message type etc.
+      snippet: t.last_message.nodes[0].snippet,
+      snippetAttachments: t.last_message.nodes[0].extensible_attachment, // TODO: not sure if it works
+      snippetSender: utils.formatID((t.last_message.nodes[0].message_sender.messaging_actor.id || "").toString()),
+      lastMessageTimestamp: t.last_message.nodes[0].timestamp_precise, // timestamp in miliseconds
+      lastReadTimestamp: (t.last_read_receipt.nodes[0]?t.last_read_receipt.nodes[0].timestamp_precise:null), // timestamp in miliseconds
+      cannotReplyReason: t.cannot_reply_reason, // TODO: inspect possible values
+
+      // @Legacy
```
bsansouci: <3 those annotations, thanks

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+      // can somebody inspect it? where is it used?
+      // probably Messenger Day uses it
+      reactionsMuteMode: t.reactions_mute_mode,
+      mentionsMuteMode: t.mentions_mute_mode,
+      isArchived: t.has_viewer_archived,
+      isSubscribed: t.is_viewer_subscribed,
+      timestamp: t.updated_time_precise, // in miliseconds
+      // isCanonicalUser: t.is_canonical_neo_user, // is it always false?
+      // TODO: how about putting snippet in another object? current implementation does not handle every possibile message type etc.
+      snippet: t.last_message.nodes[0].snippet,
```
bsansouci: Can we make this a bit more robust and check the presence of these elements before accessing? (same goes for everything below)

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+  });
+}
+
+module.exports = function(defaultFuncs, api, ctx) {
+  return function getThreadList(limit, timestamp, tags, callback) {
+    if (!callback && (utils.getType(tags) === "Function" || utils.getType(tags) === "AsyncFunction")) {
+      callback = tags;
+      tags = [""];
+    }
+    if (utils.getType(limit) !== "Number" || limit%1 !== 0 || limit <= 0) {
```
bsansouci: Can you do `isNaN` instead of this modulo trick?

bsansouci: Unless this isn't to check if limit is NaN?

src/getThreadList.js
```diff
@@ -1,73 +1,211 @@
...
+      throw {error: "getThreadList: limit must be a positive integer"};
+    }
+    if (utils.getType(timestamp) !== "Null" &&
+       (utils.getType(timestamp) !== "Number" || timestamp%1 !== 0)) {
```
bsansouci: Same as above?

src/getThreadList.js
ravkr: Why is it needed? Are there any cases where thread id wouldn't exist?

src/getThreadList.js
ravkr: For everything? :/
Why not only for these with nodes[0]?

src/getThreadList.js
ravkr: This is to check if the number is an integer and not a floating point number

src/getThreadList.js
ravkr: Yup, it also looks for an integer

src/getThreadList.js
forcemagic: How about `Number.isInteger()`? (See [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger))

forcemagic: How about `Number.isInteger()`? (See [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger))

src/getThreadList.js
ravkr: AFAIK on Facebook almost everything have its ID 😢 
but I'll add it

src/getThreadList.js
ravkr: good idea

src/getThreadList.js
ravkr: done

ravkr:
I think we can merge it now

ravkr:
@bsansouci @Schmavery ?

ravkr:
*two months later*

bsansouci:
Done! Thanks for making this and keeping it up :)

PR #577, opened 2018-01-27
Rewrite presence

### LINKED ISSUE #549: Reliable friend monitoring.
So am I missing it or is there currently no way of properly maintaining a consistent list of online friends?

using the api.listen with updatePresence option on,  I get two statuses: 0 for offline and 2 for online.
Problem is, about every 2 minutes, facebook sends an "update", and if I'm understanding it correctly, it's using 0 to indicate that the user's status hasn't changed.  

That makes it near impossible, as far as I can tell, to reliably monitor who is classified as 'online' or 'offline'.

Am I missing something?

forcemagic:
Alright, I think this is now ready. I thought it through many times, and this is the easiest way to reduce the noisiness of the `proxy-presence` thing. If it contains an update (aka a _status code_), it fires, if not, then it won't. What does everyone think about this?

(Fixes #549)

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -947,7 +947,7 @@ The message object will contain different fields based on its type (as determine
 	</tr>
 	<tr>
 		<td><code>timestamp</code></td>
-		<td>How old the presence information is.</td>
+		<td>The time when the user was last online.</td>
 	</tr>
 	<tr>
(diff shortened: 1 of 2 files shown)


Schmavery:
Thanks @forcemagic :)
I hope this helps people dealing with presence

PR #579, opened 2018-02-01
Update doc id for getThreadInfoGraphQL (#578)

### LINKED ISSUE #578: Errors in getThreadInfoGraphQL
`getThreadInfoGraphQL` is currently broken for me, failing with

```
ERR! getThreadInfoGraphQL Error: well darn there was an error_result
```

With a little digging, it can be traced back to [this line](https://github.com/Schmavery/facebook-chat-api/blob/513fb77aa635e46e6d2621e0763cc5d6a13957ce/src/getThreadInfoGraphQL.js#L99), which seems to check whether the backend is indicating a failure and bail out if it is.

The response payload for this is in the form

```js
Array(2) [Object, Object]
length: 2
0: Object {o0: Object}
1: Object {successful_results: 0, error_results: 1, skipped_results: 0}
```
As you can see, the payload is reporting an error_result and no successful_results despite the data being successfully obtained from the backend (the first item in the array contains valid info about the thread).

A monkey patch for this is disabling the error-checking to allow the response through without throwing the error. This obviously isn't a permanent solution, but I'm not quite sure why the response is reporting an error_result even though it successfully retrieved the information.

AstroCB:
Fixes #578

### DIFF
--- a/src/getThreadInfoGraphQL.js
+++ b/src/getThreadInfoGraphQL.js
@@ -72,8 +72,8 @@ module.exports = function(defaultFuncs, api, ctx) {
     var form = {
       "queries": JSON.stringify({
         "o0":{
-          // This doc_id was valid on July 15th 2017.
-          "doc_id":"1527774147243246",
+          // This doc_id is valid as of February 1st, 2018.
+          "doc_id":"1498317363570230",
           "query_params":{
             "id": threadID,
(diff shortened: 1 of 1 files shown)


bsansouci:
Thanks for finding this! I’m really surprised we have to do this :(

PR #601, opened 2018-03-06
Fix some errors on getThreadHistoryGraphQL

bruno-cadorette:
Hello,

I recently used getThreadHistoryGraphQL to export my 95 000 messages group chat on facebook. I got some issues regarding this function so I fixed them and I was able to get every messages.

### DIFF
--- a/src/getThreadHistoryGraphQL.js
+++ b/src/getThreadHistoryGraphQL.js
@@ -123,9 +123,9 @@ function extractThumbnail(media){
     }
   }
   return {
-    uri : "",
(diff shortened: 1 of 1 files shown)


ravkr:
I think that it would be better to return `null`s instead of empty strings

bruno-cadorette:
Done!

Schmavery:
This code has been recently reworked to be a little more consistent. Thanks a lot for the contribution and sorry we didn't get to merge it sooner!

PR #604, opened 2018-03-11
Attempt to fix redirection to certain Facebook subdomains

### LINKED ISSUE #592: Can't send message
I have tried the echo bot example and it is thrown an error when I send a message to the bot.

The bot didn't reply my message and here is the error.
asd

and the code
qweasd

Toohpick:
Fixes #592 
Please let me know if there is a better way to do this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -316,6 +316,17 @@ function loginHelper(appState, email, password, globalOptions, callback) {
 
   mainPromise = mainPromise
     .then(function(res) {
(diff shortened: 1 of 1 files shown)


Schmavery:
can anyone without this redirect issue verify that the normal path still works as expected? (@ravkr )

bsansouci:
Hope this works for people having this issue! Thanks for the fix :)

ALiangLiang:
I just tested the old and this version. Both can work. 👍

Schmavery:
Great, thanks @ALiangLiang

Issue #599, opened 2018-03-03
use coding style?


ALiangLiang:
Do we use any coding style? If we use some style like [standard](https://github.com/standard/standard) or [airbnb](https://github.com/airbnb/javascript), the code will look more comfortable. I suggest use [standard](https://github.com/standard/standard) on this project. And add a .eslintrc file to help developer to lint code.

Schmavery:
Is there some pattern that you think is making it hard to contribute?

ALiangLiang:
- Adding [.eslintrc](https://eslint.org/docs/user-guide/configuring) is useful for auto-lint code. Developer do no need to create eslint configuration file themeselves.
- It makes it easy to unify code style and make it beautiful.
- Some contributers [dosen't follow exactly camel case](https://github.com/Schmavery/facebook-chat-api/pull/556/files#diff-f05127a8eefa1bf9353a037143b591dbR217). Let some variables or object key like "foo_bar", some like "fooBar". This will confuse the user.
- Most large projects also use eslintrc to unify code style.

Schmavery:
I'm not entirely against the idea of aiming for some sort of unified style, though I'm not that into no semicolons (as standard appears to be).
Maybe something like prettier? Downside of having to touch every file and adding another step to development (would also be required of linting), upside of no manual (and error-prone) work to change and maintain style. I guess you would maybe still want a linter to enforce camelCase for instance. Would be nice to have a check for that.
@bsansouci thoughts?

ALiangLiang:
We can just create a lint configure file (.eslintrc). And let developers include that into editor by themselves.
Or you can also use "husky"+"lint-stage", to force developer auto lint codes before commit.

ALiangLiang:
[ref](https://hackernoon.com/what-javascript-code-style-is-the-most-popular-5a3f5bec1f6f)

Schmavery:
Thanks for the link @ALiangLiang, looks like my preference for with-semicolon style is justified by science 😜 

Looks like @ravkr created https://github.com/Schmavery/facebook-chat-api/pull/611 to try out some changes. How do people feel about the airbnb style guide? From looking at it briefly it seems to match most closely what we're doing here already (in terms of semicolons, indentation and use of single-quotes for strings).
@bsansouci

ALiangLiang:
I agree. airbnb is more like current coding style. But... Airbnb style is very strict and annoying. QQ
[...]

PR #611, opened 2018-03-14
ESLint formatting

ravkr:
I've put some ESLint rules as proposed in #599 and used automated fixing on the code (there are few more things to review by hand), if you think that something would look better in another way, just propose it here

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,48 @@
+module.exports = {
+    "env": {
+        "es6": true,
+        "node": true
+    },
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaFeatures": {
+            "jsx": true
+        },
+        "sourceType": "module"
+    },
+    "rules": {
+        "indent": [
+            "error",
+            2,
+            { "SwitchCase": 1 }
+        ],
+        "linebreak-style": [
+            "error",
+            "windows"
+        ],
+        "quotes": [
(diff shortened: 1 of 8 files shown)


Schmavery:
Thanks @ravkr, helps to be able to visualize it. Do these eslint rules match one of the standards mentioned in #599?

ravkr:
I didn't check every aspect of these standards
I've just installed eslint and then just put some rules to more of less match current style
eg. object should look like this:
```js
var obj = {
  abc: "hello"
}
```
like lines 121 and 125 here
obraz

IMO neither *standard* nor *airbnb* fits us, but this one is closer to *airbnb*

ALiangLiang:
Maybe we can use standard style. If @Schmavery really like semicolons. We can use [this style](https://github.com/Flet/semistandard) based on standard but use semicolons. It suitable for semicolons fans.

ravkr:
Why not make list with our own rules? Like the one in this commit (with few more of course)

ALiangLiang:
um..... I think this is a big project which contributing by many developers. They have different habit about coding style. So I suggeted use some famous popular coding style is better.

Of course, we need to choose most like us style.

I have some reasons:
1. This project has different style now.
2. ~~Apply popular style is easy 😛~~

Schmavery:
Going to close this because of the prettier changes. Thanks for the feedback!

PR #622, opened 2018-03-21
add .eslint.js

ALiangLiang:
1. add .eslint.js used to let developer easy to set up environment.
2. CLI: use eslint instead of prettire. It's accepted by many developer.
3. "lint" script is many developer's habit.

This commit is not useful for "use". So you don't need to merge it before release new version.


bsansouci:
No sorry we're going to go ahead with Prettier. Thanks

ALiangLiang:
haha. Actually I don't mind any kind of coding style. This PR is still use "prettier" not other style. Just replace linter "prettire" by "eslint".

PR #627, opened 2018-03-27
Fix getThreadHistory to always return attachments as an array

ivankolesnikov:


Schmavery:
Looks good, thanks for catching.

Issue #632, opened 2018-03-28
getUserID Problem


Theblood:
Hello,
my problem is this, help me please 
> ERR! getUserID TypeError: Cannot read property 'entries' of undefined
> ERR! getUserID     at /root/olaf-bot/node_modules/facebook-chat-api/src/getUserID.js:43:35
> ERR! getUserID     at tryCatcher (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:26:23)
> ERR! getUserID     at Promise._settlePromiseFromHandler (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:510:31)
> ERR! getUserID     at Promise._settlePromiseAt (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:584:18)
> ERR! getUserID     at Promise._settlePromises (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:700:14)
> ERR! getUserID     at Async._drainQueue (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:123:16)
> ERR! getUserID     at Async._drainQueues (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:133:10)
> ERR! getUserID     at Immediate.Async.drainQueues (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/async.js:15:14)
> ERR! getUserID     at runCallback (timers.js:672:20)
> ERR! getUserID     at tryOnImmediate (timers.js:645:5)
> ERR! getUserID     at processImmediate [as _immediateCallback] (timers.js:617:5)
> ERR! getUserID  TypeError: Cannot read property 'entries' of undefined
> ERR! getUserID     at /root/olaf-bot/node_modules/facebook-chat-api/src/getUserID.js:43:35
> ERR! getUserID     at tryCatcher (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/util.js:26:23)
> ERR! getUserID     at Promise._settlePromiseFromHandler (/root/olaf-bot/node_modules/facebook-chat-api/node_modules/bluebird/js/main/promise.js:510:31)
[...]

PR #633, opened 2018-03-29
Rename domain www.facebook.com to web.facebook.com

Theblood:
for changes of facebook, this is  minimal change, is solution to problem, when execute getUserID.

### DIFF
--- a/src/getUserID.js
+++ b/src/getUserID.js
@@ -33,7 +33,7 @@ module.exports = function(defaultFuncs, api, ctx) {
     };
 
     defaultFuncs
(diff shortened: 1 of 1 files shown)


Schmavery:
Unless there are objections, let's go with the approach I suggested here? https://github.com/Schmavery/facebook-chat-api/issues/632#issuecomment-377293267

Theblood:
@Schmavery @ravkr  I would have to see how many redirects there may be

Schmavery:
Thanks for investigating this issue @Theblood. I'm going to close this as the fix here won't work for everyone, let's keep discussing the solution in #632 if necessary.

PR #636, opened 2018-04-07
Fix blob attachment variable in getThreadHistory

### LINKED ISSUE #635: blob is not defined
Hi,

When I use the `getThreadHistory` method, I get the error `ReferenceError: blob is not defined`.
It seems that this error is only triggered when I'm getting a message containing a gif.
The error is in `getThreadHistory` at line 65 when accessing the blob variable, which seems to never be defined.

Thank you for your help.

Toohpick:
Fixes #635


Schmavery:
Thanks! (Would have merged sooner but was out of town for the week and didn't see)

PR #638, opened 2018-04-27
Moved prettier to dev-dependencies, fixed scripts

Bjornskjald:
(saves 7,46MB of download on `npm install`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "1.5.1",
   "description": "Facebook chat API that doesn't rely on XMPP.  Will NOT be deprecated April 30th 2015.",
   "scripts": {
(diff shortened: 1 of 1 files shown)


forcemagic:
Yep, in an npm script, `node_modules/.bin` is automatically appended to `$PATH`, so I can totally agree with @Schmavery

Bjornskjald:
Didn't know it, thanks.
I thought if mocha is with `node_modules/.bin` then I could change prettier too, just for consistency

forcemagic:
The tests were added 3 years ago (at least that's what blame shows), when this feature was not present (I guess), but referring to mocha's exact path is also unnecessary.

PR #649, opened 2018-05-27
Allow setting User Agent using an environment variable

choyg:
Facebook uses user agent as a factor when deciding if a login attempt is legitimate or not.

If you use facebook-chat-api on a hosted server (gcp, aws, azure, heroku, etc.) there is a high chance your login attempt gets blocked (this is from my experience) and account temporarily locked. However, by using your real user agent string FB tends to recognize chat-api as legitimate requests.

This PR allows someone to specify a custom user-agent if they wish otherwise defaults to the Mac OS X one already in this codebase.

EDIT: It might be more friendly to make this a login param instead?

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -885,7 +885,7 @@ function generateTimestampRelative() {
   return d.getHours() + ":" + padZeros(d.getMinutes());
 }
 
-function makeDefaults(html, userID, ctx, options) {
+function makeDefaults(html, userID, ctx) {
   var reqCounter = 1;
   var fb_dtsg = getFrom(html, 'name="fb_dtsg" value="', '"');
 
@@ -957,11 +957,11 @@ function makeDefaults(html, userID, ctx, options) {
   }
 
   function postWithDefaults(url, jar, form) {
-    return post(url, jar, mergeWithDefaults(form), options);
+    return post(url, jar, mergeWithDefaults(form), ctx.globalOptions);
   }
 
   function getWithDefaults(url, jar, qs) {
-    return get(url, jar, mergeWithDefaults(qs), options);
+    return get(url, jar, mergeWithDefaults(qs), ctx.globalOptions);
   }
 
   function postFormDataWithDefault(url, jar, form, qs) {
@@ -970,7 +970,7 @@ function makeDefaults(html, userID, ctx, options) {
       jar,
       mergeWithDefaults(form),
       mergeWithDefaults(qs),
-      options
+      ctx.globalOptions 
     );
   }
 
@@ -1277,3 +1277,4 @@ module.exports = {
   getAppState,
   getAdminTextMessageType
 };
+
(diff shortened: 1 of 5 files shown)


Schmavery:
Seems like it would make most sense to make this a value settable using setOptions

ravkr:
using environment variables is probably good idea, but I think that we should be able to use setOptions, or even better: `login({email: "...", password: "...", userAgent: "Firefox/62.0"}, ...)`

choyg:
Agreed this should go in setOptions. Didn't notice that method. Closing and will open a new PR later

ravkr:
You could just make changes and push your commits to github
Closing and opening new request is not necessary 

And there is one important thing: we need to set other browser BEFORE logging in, so setOptions probably would cause even more trouble (same session? Different browser? Is it hacker?)

Schmavery:
You can pass options during login

Schmavery:
@choyg you should be able to add support here: 
https://github.com/Schmavery/facebook-chat-api/blob/85b79ac96938c5610a82c1e8d763e2299b2b6ad7/index.js#L10-L41

And then add a default value here:
https://github.com/Schmavery/facebook-chat-api/blob/85b79ac96938c5610a82c1e8d763e2299b2b6ad7/index.js#L452-L458

choyg:
Thanks, I'll look into it

choyg:
Utils currently doesn't take an options object, should I add one for get,
post, and other functions which use getHeaders? It'd require modifying all
existing usage of get/post but that seems the most inline with how the
codebase currently passes objects like jar, options and ctx

Schmavery:
It would be easier to answer the question if we had some idea of how often those functions get used and how many spots would have to be changed, but yeah that's likely the right idea.

choyg:
I was able to add an options object to the base get, post, and postFormData functions in utils like so:
https://github.com/choyg/facebook-chat-api/blob/468b580b5e88f972213e6bf4787588701e6ba5f6/utils.js#L959-L965

This way we don't have to change any of the API calls since the options will be automatically included. I did have to fix a few of the API functions to not call utils.get() directly but instead use the built API (defaultFuncs).

I tried to run the test suite but got 8 failing tests even with no changes? Seems like it's not updated since there's some old graphql suffixed functions.

utils.js
```diff
@@ -886,7 +885,7 @@ function generateTimestampRelative() {
   return d.getHours() + ":" + padZeros(d.getMinutes());
 }
 
-function makeDefaults(html, userID, ctx) {
+function makeDefaults(html, userID, ctx, options) {
```
Schmavery: I believe `options` is equivalent to `ctx.globalOptions`

choyg:
Thanks, changed to use ctx.

Listen works on my end, I tried with my program and it worked fine but another data point is always nice.

Schmavery:
thanks

PR #651, opened 2018-06-03
Fix support for page listener

ngxson:
**Facebook is now killing Messenger Platform for Pages!**

Now apps want to be reviewed must be associated to a business. Sadly, they will ask you to provide documents to prove that your business is real and is making money.

With this change we can use this api instead of the dead one provided by Facebook.

Please merge this change, so the poor like me can have some hope.. and not something like Chatfuel which has a very limited range of features.

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -32,8 +32,14 @@ module.exports = function(defaultFuncs, api, ctx) {
     state: "active",
     idle: 0,
     cap: "8",
-    msgs_recv: msgsRecv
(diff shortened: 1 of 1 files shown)


ravkr:
"Please merge" you don't have to ask us to merge... just make sure that everything works, and after some people verify it, we will merge
anyway... screw you Zuckerberg :/

bsansouci:
@ravkr hey hey let’s not be aggressive here. 
Seems good to me, I’m trusting you that this works.

PR #653, opened 2018-06-07
Add eslint error checking w/ travis

Schmavery:
Maybe this can help catch some simple bugs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: node_js
+node_js:
+  - "6"
+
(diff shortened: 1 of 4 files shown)


ravkr:
yeah... eslint is a good idea #611

Schmavery:
Sorry @ravkr, didn't mean to minimize your work. I think everything got mixed up with formatting etc.  We wanted to not have the linter complaining about style when we could just reformat easily :)

Issue #624, opened 2018-03-23
Investigate and add admin toggling capabilities


tarungog:
New messenger features allow users to remove and add admins in messenger group chats

ravkr:
New getThreadList have this commented out, I will change it when I get back home

AstroCB:
This also ties into membership approval requests, which the API may be able to field.

In my testing, it seems the API cannot freely add people as before when admin approval for membership is turned on, even if the controlling account is an admin, so it would be nice if we could add functionality to approve/deny these requests as needed.

Schmavery:
I think this was added in https://github.com/Schmavery/facebook-chat-api/pull/659?? If there's some issue with it we can reopen. Thanks all!

PR #658, opened 2018-06-14
Add changeAdminStatus

AstroCB:
Adds a `changeAdminStatus` function to allow the API to alter the admin status of users in group chats (naming suggestions welcome).

re: #624

### DIFF
--- a/src/changeAdminStatus.js
+++ b/src/changeAdminStatus.js
@@ -1,0 +1,78 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+  return function changeAdminStatus(userIDs, makeAdmin, threadID, callback) {
+    if (
+      !callback &&
+      (utils.getType(makeAdmin) === "AsyncFunction" ||
+        utils.getType(makeAdmin) === "AsyncFunction" ||
+        utils.getType(threadID) === "Function" ||
+        utils.getType(threadID) === "AsyncFunction")
+    ) {
+      throw { error: "please pass makeAdmin and threadID as the second/third arguments." };
+    }
+
+    if (utils.getType(userIDs) === "String") {
+      userIDs = [userIDs];
+    }
+
+    if (!callback) {
+      callback = function() {};
+    }
+
+    var messageAndOTID = utils.generateOfflineThreadingID();
+    var form = {
+      client: "mercury",
+      action_type: "ma-type:log-message",
+      author: "fbid:" + ctx.userID,
+      author_email: "",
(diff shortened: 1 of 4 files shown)


ravkr:
could you do
```bash
git remote add upstream https://github.com/Schmavery/facebook-chat-api.git
git rebase upstream/master
git push -f
```
it would get rid of all these duplicated commits in every pull request :/

AstroCB:
Good call – thanks @ravkr

src/changeAdminStatus.js
```diff
@@ -0,0 +1,82 @@
...
+module.exports = function(defaultFuncs, api, ctx) {
+  return function changeAdminStatus(userIDs, makeAdmin, threadID, callback) {
+    if (
+      !callback &&
+      (utils.getType(makeAdmin) === "AsyncFunction" ||
+        utils.getType(makeAdmin) === "AsyncFunction" ||
+        utils.getType(threadID) === "Function" ||
+        utils.getType(threadID) === "AsyncFunction")
+    ) {
+      throw { error: "please pass makeAdmin and threadID as the second/third arguments." };
```
Schmavery: If these are required fields, we can probably just check their types rather than checking if they're not a callback?

Schmavery:
Thanks @AstroCB 
@ravkr does this PR do the same thing as #659? What's your concern with this one? I don't see much difference other than extra fields in the form, can someone confirm whether these are sent by the browser? If so, we should mimic that.
Or is there something else I'm missing?

ravkr:
Yes, both PR are doing the same thing
But mine looks a bit cleaner (no extra fields) 
I copied that request from messenger.com

Schmavery:
@AstroCB did you find these fields are required/used by the browser when you looked or do you think we can just remove them?

AstroCB:
Nope, they’re just holdovers from another file that I used as the basis for this – should be good to remove.

src/changeAdminStatus.js
AstroCB: Yep, another holdover from the basis file I used.

Schmavery:
@ravkr 's change has no docs though 😛
At this point idk which one to spend time reviewing, if you two want to decide if/how you want to split this up, I don't see a problem with adding this functionality.

ravkr:
You can review mine PR
I'll add docs soon

AstroCB:
Made the discussed changes last night and forgot to publish – they're up now, but either way works for me.

ravkr:
@Schmavery what's the decision?

Schmavery:
Thanks for helping out @AstroCB, we ended up merging @ravkr's branch <3

PR #659, opened 2018-06-14
Add api.changeAdminStatus

ravkr:
few days ago I've made similar function to #658
But I didn't have time to write docs so I didn't pushed it to Github
It looks a bit cleaner IMO

I've also done some tests in the wild
and you can change `adminStatus` of every user in thread without being blocked by Facebook

but if you try to do the same thing one by one... Facebook blocks you for an hour :(

obraz

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -2,6 +2,7 @@
 
 * [`login`](#login)
 * [`api.addUserToGroup`](#addUserToGroup)
+* [`api.changeAdminStatus`](#changeAdminStatus)
 * [`api.changeArchivedStatus`](#changeArchivedStatus)
 * [`api.changeBlockedStatus`](#changeBlockedStatus)
 * [`api.changeGroupImage`](#changeGroupImage)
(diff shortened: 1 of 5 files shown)


ravkr:
changelog record and docs added

ravkr:
@bsansouci, could we merge?

DOCS.md
```diff
@@ -22,7 +23,6 @@
 * [`api.getThreadPictures`](#getThreadPictures)
 * [`api.getUserID`](#getUserID)
 * [`api.getUserInfo`](#getUserInfo)
-* [`api.threadColors`](#threadColors)
```
bsansouci: Can you undo this change? (You moved the function down)

index.js
```diff
@@ -75,17 +75,18 @@ function buildAPI(globalOptions, html, jar) {
     },
   };
 
-  var apiFuncNames = [
+  const apiFuncNames = [
     'addUserToGroup',
     'changeArchivedStatus',
     'changeBlockedStatus',
     'changeGroupImage',
+    'changeNickname',
```
bsansouci: Can you undo these moves?

ravkr:
These little changes were just sorting alphabetically
Should I really undo them?

Schmavery:
Alright had a chance to look at this again. If this is the one that you guys want to merge, let's change the name to `changeAdminStatus` to be more consistent with the other function names and then I'm happy to merge. Idrc about the function order, sorry for the delay here but in the future it's probably simplest to keep it to one PR per feature when possible.

Schmavery:
thanks!

PR #661, opened 2018-06-15
Fix d.message is null

mrcyclo:


Schmavery:
thanks for the fix :)

PR #672, opened 2018-08-03
Remove switch..case resData.error

Hongarc:
You don't need check all case of `resData.error`.


Schmavery:
Doesn't this result in worse error messages?

Hongarc:
Message will auto change my language and you don't need test all case and write it to your code.
(message from facebook)

PR #684, opened 2018-09-21
Login with access_token

t-rekttt:


ravkr:
Any description? Comments? README update? CHANGELOG update?

t-rekttt:
> Any description? Comments? README update? CHANGELOG update?

Hmm I don't know what to put yet. Maybe a demo on how to login with access_token?

ravkr:
You can just edit Readme and put there some description and example usage

ravkr:
I think we need some error handling here...
obraz

and how can I get a working token?

t-rekttt:
The exploit was related to this, and now Facebook have removed the access token in user profile page in order to fix that.
It means we have to get the access token from somewhere else. I'll try to find it later.

t-rekttt:
You could get an access_token by going to this link https://b-graph.facebook.com/auth/login?password=<password>&email=<email>&access_token=350685531728|62f8ce9f74b12f84c123cc23437a4a32&method=POST
(replace <email> and <password> with your Facebook email and password and urlencode if there is any special character). If you paste this link into a previously authorized browser, it won't ask for OTP, else you just need to replace OTP into the <password> field.
You only need to input your credentials once to generate a forever-valid access_token (unless your account run into a checkpoint, session logout or password change)

Axel186:
The approach is working?

I am getting the next error:
```
REST API is deprecated for versions v2.1 and higher (12)
```

Axel186:
OK, I think I understand how it works now.

It didn't work with my own app, but did work with yours (`350685531728|62f8ce9f74b12f84c123cc23437a4a32`), so I need to review the permissions settings to get it working.

By the way, there is a hardcoded `new_app_id: 256002347743983` in `exchangeSession` function. I assume we need to have an option to provide our own `app_id`.

t-rekttt:
The app_id is from the app Facebook Messenger for Android, one of a few apps that provide access_token that allows you to interact with Facebook messages. Replace the new_app_id with your own app it won't make it work anyway.

ravkr:
I think we need to update Readme after latest Facebook change

Schmavery:
If I understand this correctly (letting you get all the normal functionality without requiring username and password), it seems very cool. Looks like the instructions for getting the token don't work anymore (as ravkr pointed out).

t-rekttt:
> You could get an access_token by going to this link https://b-graph.facebook.com/auth/login?password=&email=&access_token=350685531728|62f8ce9f74b12f84c123cc23437a4a32&method=POST
> (replace and with your Facebook email and password and urlencode if there is any special character). If you paste this link into a previously authorized browser, it won't ask for OTP, else you just need to replace OTP into the field.
> You only need to input your credentials once to generate a forever-valid access_token (unless your account run into a checkpoint, session logout or password change)

@Schmavery because of recent data breaches, Facebook is changing alot, from removing the default access_token to limiting permissions we could get from different apps type (eg: Facebook page app won't be able to interact with user messages). But the method I quoted above still works.
There are 2 seperated problems here:
1. Getting a proper access_token. There are many other ways to get a proper one but I think my method is pretty easy and convenient, although it would still require username/password. If you run the method on a browser with a Facebook session, it won't ask for login approval.
2. Interacting with messages. That's what I implemented in my commits. The advantage of authenticating with access_token over username/password is that Facebook would never ask you for login approval whenever you change your IP.

NemOry:
I tried logging in using the login url method but facebook doesn't recognize the login so I get locked out. Does anybody know how to get around this?

t-rekttt:
@NemOry just need to approve the login manually then choose "Save browser", then use the url again. Else you could try on a browser that is already saved

lequanglam:
bump

PR #686, opened 2018-09-28
startEvent function added

darkcoderrises:
I have used some hacks to get this done. Could someone help to get it to the project's standards?

### DIFF
--- a/src/startEvent.js
+++ b/src/startEvent.js
@@ -1,0 +1,60 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+    // muteSecond: -1=permanent mute, 0=unmute, 60=one minute, 3600=one hour, etc.
+    return function startEvent(threadID, eventTime, eventName, callback) {
+
+        var serialize = function(obj, prefix) {
+            var str = [],
+                p;
+            for (p in obj) {
+                if (obj.hasOwnProperty(p)) {
+                    var k = prefix ? prefix + "[" + p + "]" : p,
+                        v = obj[p];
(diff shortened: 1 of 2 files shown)


src/startEvent.js
```diff
@@ -0,0 +1,60 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+    // muteSecond: -1=permanent mute, 0=unmute, 60=one minute, 3600=one hour, etc.
```
ravkr: mute?

Schmavery:
Does this still work? I don't seem to be able to create reminders in messenger anymore.
If we want to merge it, it would be great to get a brief explanation of some of these "hacks" and then obviously we'd need DOCS etc.

darkcoderrises:
As you can see in the commit, I found an url, couldn't encode it properly, so just left it there as is. Maybe that's why it isn't working for you. I was hoping that you could guide me as IDK either your or fb's standards.

src/startEvent.js
```diff
@@ -0,0 +1,60 @@
...
+            location_name: ""
+        };
+
+        var url1 = "https://www.facebook.com/ajax/eventreminder/create/?acontext=%7B%22action_history%22%3A[%7B%22surface%22%3A%22messenger_chat_tab%22%2C%22mechanism%22%3A%22messenger_composer%22%7D]%7D&" + serialize(form);
+
+        defaultFuncs
+            .post(
+                url1,
+                ctx.jar,
+                {}
```
Schmavery: What happens if you pass form here instead of serializing it yourself/putting it in the url?

PR #691, opened 2018-10-15
update cheerio version 0.19.0 to 0.22.0

ccam:
I updated the cheerio package from 0.19.0 to 0.22.0 This should get rid of the security warning on npm install. Mocha needs to be updated as well but I don't know if it will break testing.


Schmavery:
Seems ok. Thanks

PR #697, opened 2018-11-24
Use Buffer.from

Hongarc:
Buffer() is deprecated due to security and usability issues.
Use Buffer.from() methods instead.


Schmavery:
Ok, thanks

PR #700, opened 2018-12-01
Use rule prettier for eslint

Hongarc:
I know that this commit run with error with travis-ci
Because some commit don't run with `prettier`.

If accept this commit and fix your code with `prettier` once, any pull request will error if they don't run `prettier` before commit (use one style in code).


ravkr:
I think we can close it now.

Hongarc:
This repo has no active

PR #704, opened 2018-12-27
Changing all domain to www.facebook.com

lequanglam:
Change from www.messenger.com -> www.facebook.com to prevent "Not logged in" error.


kentnek:
This does fix the "Not logged in" error. Thanks so much!

Schmavery:
And it doesn't break any of the apis that you switched the URL for? Originally those apis were only supported on messenger.

Schmavery:
Thanks all!

PR #710, opened 2019-02-04
Way of testing the bots without creating another account

gamelaster:
Hello,

I find out that Facebook offers [Whitehat accounts](https://www.facebook.com/whitehat/accounts/), which are used for investigating security vulnerabilities. They're pretty great, it's absolutely same as normal Facebook account, just it's isolated, and you can't get blocked. I tested a facebook-chat-api with those accounts, it's works greatly, just it's little bit slower (but maybe that's because of my internet)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ If you want to use bleeding edge (directly from github) to test new features or 
 npm install Schmavery/facebook-chat-api
 ```
 
+## Testing your bots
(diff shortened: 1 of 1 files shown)


lequanglam:
@Schmavery should merge this.

Schmavery:
Great find, sorry, I didn't realize this was just a docs change :)
I'll look into this to set up CI or something when I get a chance.

PR #711, opened 2019-02-09
Unsend message feature.

lequanglam:
Facebook was updated with a new feature called "Unsend message". So here is my implentation for that feature.

Sorry if my English sucks. English is not my main language.


Schmavery:
This looks great, thanks for contributing.

PR #714, opened 2019-02-13
apiFuncNames changes.

lequanglam:


Schmavery:
I'd rather not rely on the directory query right now if we can avoid it, though it's a clever idea. I'll merge #717 instead. Thanks!

PR #717, opened 2019-03-02
Added unsendMessage to apiFuncNames in index.js

DentFuse:
Added `unsendMessage `to `apiFuncNames`, not doing this results in a `TypeError: api.unsendMessage is not a function`


lequanglam:
thanks for that, but where is @Schmavery?

Schmavery:
sorry

PR #721, opened 2019-04-13
Reply message feature & Some bug fixes

lequanglam:
took 1 week to make this, but this feature is so easy to make
kill me plz

EDIT: also including #698, that pr is getting old


7osCraft:
Attachments need to be formatted correctly

Schmavery:
Is there a way to add this without copying most of sendMessage?
In the past, this sort of duplication has been a sign that maybe it should just be integrated as a different type of message in sendMessage.

lequanglam:
sorry for late response

PixelHir:
bump

acenturyandabit:
Additional bump :)

lequanglam:
@Schmavery

PixelHir:
@Schmavery please, check this out
I had to create a forked version of the repo just to get the new stuff because this one is way too old

lequanglam:
@Schmavery bump

src/sendMessage.js
```diff
@@ -303,7 +303,7 @@ module.exports = function(defaultFuncs, api, ctx) {
     cb();
   }
 
-  return function sendMessage(msg, threadID, callback) {
+  return function sendMessage(msg, threadID, callback, replyToMessage) {
```
Schmavery: The callback is supposed to be optional, but this makes it so that you can't pass a reply without a callback.

DOCS.md
```diff
@@ -1372,7 +1428,7 @@ __Arguments__
 ---------------------------------------
 
 <a name="sendMessage"></a>
-### api.sendMessage(message, threadID[, callback])
+### api.sendMessage(message, threadID[, callback[, messageID]])
```
Schmavery: What does it mean for the [[]] to be nested? That you can't pass messageId without callback? Why?

lequanglam:
ok, fixed that

Issue #724, opened 2019-04-19
api.markAsRead does not work with PageID


Pavelas:
Hello,

I was trying to use the bleeding edge version of facebook-chat-api with PageID. I found the problem, when you are trying to **markAsRead()** messages and **PageID is defined**, there is an error:

`errorSummary: 'This content is no longer available',`

It seems that **markAsRead()** request for Facebook pages is different.

PR #725, opened 2019-04-19
Fix api.markAsRead() for Facebook pages

Pavelas:
Issue #724

### DIFF
--- a/src/getThreadHistory.js
+++ b/src/getThreadHistory.js
@@ -592,6 +592,7 @@ module.exports = function(defaultFuncs, api, ctx) {
     // `queries` has to be a string. I couldn't tell from the dev console. This
     // took me a really long time to figure out. I deserve a cookie for this.
     var form = {
+      "av": ctx.globalOptions.pageID,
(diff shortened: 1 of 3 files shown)


Schmavery:
Does it work to unconditionally include the extra field for getThreadHistory and getThreadList?
Generally it's better to use the official api for pages but I don't mind including this if it doesn't break regular usage.

Pavelas:
Yes it works unconditionally, I made some tests with my personal account and facebook page as well.

Official API is so messy, just to read incoming messages from facebook page, you have to create application, submit for review, generate tons of tokens, connect your facebook page, etc...

Also I found a small bug api.listen(), even if you have pageID in globalOptions, it still listen some events from your personal messages, such as read_receipt and message_reaction, but it is not critical.

Schmavery:
Feel free to submit fixes for page-related bugs, I don't use it so it's more annoying to test for me. As long as you can confirm they don't break non-page usage, I'm happy to merge.
Thanks!

PR #648, opened 2018-05-23
Add getLastNotifications Command (#644)

ezaca:
In issue #644 I was discussing a personal need I had. Now I have made a command to fit my needs: it makes a request and gets a list of the last notifications of the currently logged user.

As it is not a Messenger thing, but an FB thing, I don't know if it's expected to be in the official Unofficial Facebook Chat API. I will maintain a fork for a while and use it from that, if you like the function and think it can be somewhat useful, I am opening a pull request.

The initial commit includes source, documentation, and test for the getLastNotification command.

If you are willing to accept the pull request, please review the additions and inform me of any problem related to code, formatting or grammar (English is not my native language).

Issue #644, opened 2018-05-19
Read notifications of Facebook profile?


ezaca:
Hi, everyone!

I was playing around with this great library, and I was trying to list notifications of a Facebook profile (not Messenger).

**First**, i don't know if this library is intended for that, but it seems to work and I would love to use it for a personal application that informs me about notifications in more than one account in same browser (shhh, don't let they know).

**What did I do? Step-by-step**

- Install the module or clone this repo;
- Add a file `module/src/getNotificationList.js` and copy from `getThreadList.js`;
- Register it in `index.js`;
- Remove all argument checks, and all arguments, except `callback`;
- Call `defaultFuncs.post` with our form (I cleared the form and added only the "data" and "__a" field, see below for details);
- Make a test file with the first example in the docs, but replace the `listen` function call by our method;
- Run the test and get the answer. Do not look for the `payload` as the getThreadList do, but get the whole answer and look for a `jsmods` field.

**What is wrong?**
I got the answer, it's a JSON, it's OK, but it's pretty different from what I get in the browser! In Firefox I get all the fields with the notifications, but in Node I get an incomplete answer. The field with the notifications is missing.

**What is the request that works in browsers?**
With a logged account, access [this link](https://www.facebook.com/ajax/pagelet/generic.php/WebNotificationsPayloadPagelet?data={%22businessUserID%22:null,%22length%22:10,%22clientRequestID%22:%22js_3u%22}&__a=1).
The link is composed by:
- the URL of `facebook.com`;
- the URI to `ajax/pagelet/generic.php/WebNotificationsPayloadPagelet`
- a `data` JSON with `{businessUserID, length, clientRequestID}`;
- and a `__a` field.

The fields are OK, but you must send the cookies too. Because of that, I tried to use the `post` function matching what getThreadList did.
In Firefox, for example, I can find (Ctrl+F) for the text of my last notifications, but they do not appear on Node.

So what I want is to receive the same JSON as I received in Firefox, so I can access `response.jsmods.require[0][3][2].nodes` array (this index will be resolved programatically, not hard-coded) and get the information about the last notifications in the profile.

Any hint of what is going on? Did the link worked to you?

bsansouci:
Cool stuff!
So you'll need to really replicate the request the browser makes, down to all the headers, all the query parameters and all data you send as body (when it's a POST).
You said you cleared the form, make sure you didn't remove anything, because those might be used (well you never know).
You'll also need to make the correct type of request GET/POST.
[...]

PR #728, opened 2019-06-11
Added getNotifications to the API and updated documentation

kueblc:
Meets needs as in #644. This is a slightly different implementation and endpoint than #648 and I noticed #648 has been outstanding for a while.


lequanglam:
@Schmavery bump

PR #738, opened 2019-07-16
added reply listening

acenturyandabit:
It probably needs a bit more work and some testing, but here's a proof-of-concept :)


lequanglam:
oof, already add that features in #721 lol
please look at PR tab to see if your upcoming PR is duplicated with other PRs.

acenturyandabit:
#721 doesnt seem to have reply _listening_. You have reply sending (onya, i used your code) but not reply listening which i added :)

lequanglam:
well, look like you're wrong, please look at file src/listen.js from line 319 to line 383.

acenturyandabit:
I stand corrected. I would have liked replies to passed through the normal message pathway instead of having a separate 'message-reply' type, but I can live with that. Cheers.

PR #739, opened 2019-07-25
[docs] Define obj variable in docs for login with login approvals/2fa

rohan-varma:
The current docs for [login](https://github.com/Schmavery/facebook-chat-api/blob/master/DOCS.md#login) have a slight error in the code for login approvals/2 factor auth. In the code example, a variable `obj` is used, but it is not defined, resulting in a `ReferenceError`. This diff adds a definition for `obj` in the code example.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -137,6 +137,7 @@ var rl = readline.createInterface({
   output: process.stdout
 });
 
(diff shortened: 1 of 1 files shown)


Schmavery:
thanks

PR #747, opened 2019-08-24
Bug fix for reply message feature.

lequanglam:
Yeah, sometimes Facebook doesn't wrap things in blob_attachment like this:
image
This bug fix will resolve the error I mentioned here.
How to reproduce this bug: Just reply to a message that contains stickers.

@Schmavery sorry for that looooong compare, but pls merge fast


Schmavery:
thanks

PR #750, opened 2019-09-05
Fix incorrect error assumption in sendMessage

xnerhu:
Hello. If I use `api.sendMessage` without the last argument (reply message id), I get `MessageID should be of type string and not String.` So I fixed it in this PR.


lequanglam:
Thank you for the fix and sorry for that bug i made.
Verified that it works.
@Schmavery

Schmavery:
thanks

PR #756, opened 2019-10-18
Added support to detect changes in polls

MartenBE:
This PR would allow the users of the [matrix facebook puppet bridge](https://github.com/matrix-hacks/matrix-puppet-facebook) to detect changes in polls.

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -411,6 +411,20 @@ module.exports = function(defaultFuncs, api, ctx) {
                         case "change_thread_nickname":
                         case "change_thread_icon":
                           break;
+                        case "group_poll":
+                          var fmtMsg;
+                          try {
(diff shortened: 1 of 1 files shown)


src/listen.js
```diff
@@ -411,6 +411,20 @@ module.exports = function(defaultFuncs, api, ctx) {
...
+                              detail: err,
+                              res: v.delta,
+                              type: "parse_error"
+                            });
+                          }
+                          return globalCallback(null, fmtMsg);
+                          break;
                         default:
                           return;
                       }
```
Schmavery: Why can't we just add `case "group_poll":` to the `ThreadName`, `ParticipantsAddedToGroupThread`, `ParticipantLeftGroupThread` section right below?

Schmavery:
Seems alright to add. Thanks :)

PR #765, opened 2019-11-27
Added support for MQTT

7osCraft:
Added support for MQTT with the existing events and messages.
Also added support for marking a message as delivered because I think that causes facebook to temporarly block accounts.
Also added a "change_thread_image" event triggered when a group image is changed.

To listen with mqtt simply change ```api.listen``` to ```api.listenMqtt```


src/listenMqtt.js
```diff
@@ -0,0 +1,443 @@
+"use strict";
+var fbconnect = require("./mqtt/fbconnect");
+var utils = require("../utils");
+var log = require("npmlog");
+
+var identity = function () { };
+var lastSeqId = 0;
+
+var defaultFuncs = undefined;
```
Schmavery: Why do this rather than just passing defaultFuncs into listenMqtt?

src/listenMqtt.js
```diff
@@ -0,0 +1,443 @@
...
+				undefined :
+				globalCallback(null, formattedEvent);
+	}
+}
+
+module.exports = function (df, api, ctx) {
+	var currentlyRunning = null;
+	var globalCallback = identity;
+	defaultFuncs = df;
+	return function (callback) {
```
Schmavery: I find this function a bit confusing, but tbf it's been a while since I spent time with the old listen code. Is this the stopListening callback? it seems to do a lot of stuff, whereas the listenMqtt function itself doesn't seem to do anything.
Maybe I'm reading this wrong.

Schmavery:
Wow this is great, thanks.

Can we mark-as-delivered automatically? Generally I'd say that that and change_thread_image could have been in a separate PR though.

I left a couple questions in the code, but it generally looks like it could work.
If you can add a brief note to the docs about the listenMqtt function being an experimental way to receive messages and a small usage example, I think we can probably merge this. If it works well for people, we can transition in a while to having it be the default listen behaviour. I'm assuming the messages that we get through this are pretty much the same as the ones we used to get through the regular listen?

In the meantime, if anyone else who is excited about this functionality wants to pull the branch and try it out, that's always helpful for a big change like this. Thanks!

src/listenMqtt.js
7osCraft: Yeah sure I think that would be better.

src/listenMqtt.js
7osCraft: This is the function itself that you pass another callback function to it to recieve messages and events.
stopListening is not there yet but I'll try to add it too.

7osCraft: This is the function itself that you pass another callback function to it to recieve messages and events.
stopListening is not there yet but I'll try to add it too.

7osCraft:
Sure. We also could add an option for auto marking messages as seen.
for change_thread_image I could do that in another PR with some other events I found related to calling.

And yes, messages are exactly the same.
Thanks!

7osCraft:
Also it would be helpful to merge #628 because of ForcedFetch.

ravkr:
I'd suggest changing tabs to spaces, to match rest of the code.

7osCraft:
@Schmavery Would you like me to remove markAsDelivered and change_thread_image from this PR and do another one after this is merged?

src/listenMqtt.js
7osCraft: Added stopListening.

src/mqtt/fbws.js
```diff
@@ -0,0 +1,93 @@
...
+var urlModule = require('url')
+var WSS_OPTIONS = [
+  'rejectUnauthorized',
+  'ca',
+  'cert',
+  'key',
+  'pfx',
+  'passphrase'
+]
+var IS_BROWSER = process.title === 'browser'
```
Schmavery: Are we actually doing something with the browser or process.title?

src/mqtt/fbws.js
```diff
@@ -0,0 +1,93 @@
...
+  setDefaultOpts(opts)
+  var url = buildUrl(opts, client)
+  return websocket(url, undefined, opts.wsOptions)
+}
+
+function buildBuilder(client, opts) {
+  return createWebSocket(client, opts)
+}
+
+function buildBuilderBrowser(client, opts) {
```
Schmavery: Maybe we can remove this function? Unless I'm missing something

src/mqtt/fbws.js
```diff
@@ -0,0 +1,93 @@
+//From mqtt.js
```
Schmavery: Is this copied from somewhere? It's ok, but can you put a more full URL or something so that we can find it again later?

Schmavery:
I don't really mind change_thread_image staying but lets pull out markAsDelivered for now if you don't mind -- I think we should start trying to do that and markAsRead automatically and I don't want that to block this PR :)

Thanks for adding the docs! I left just a couple more comments and then I think we should be good to merge unless someone else has concerns.

src/mqtt/fbws.js
7osCraft: Actually no, this is copied from MQTT.js itself.
I can look into unused things like this and remove it.

src/mqtt/fbws.js
7osCraft: Sure. I will add an url to this.

7osCraft: Sure. I will add an url to this.

7osCraft:
Thank you! I'm glad I could help.
I have no problem in removing markAsDelivered for now, I can work on making it automatic with markAsRead.

PR #628 has a good way to handle GraphQL requests that way most requets are in one place.
I could try to merge it and make it up to date if you like.

I'm also thinking about refactoring parseDelta so we could add more events in the future easily. But I'll leave that for another PR.

Thanks!

src/mqtt/fbws.js
7osCraft: Removed in last commit.

7osCraft:
Could you please take a look at the latest changes?

lequanglam:
image
:|

ravkr:
@lequanglam what was in that message? was it normal text? emoji? sticker? picture? that would probably help

lequanglam:
@ravkr i can't check what messages is throwing that error (cuz there's a LOT of messages going in and out), but i hooked up some error logging, hopefully it will catch what messages is messing up.
image

lequanglam:
found, it was ForcedFetch that cause issues.
image

src/listenMqtt.js
```diff
@@ -0,0 +1,460 @@
...
+    } else if(topic === "/orca_presence") {
+      if(!ctx.globalOptions.updatePresence) {
+        for(var i in jsonMessage.list) {
+          var data = jsonMessage.list[i];
+          var userID = data["u"];
+
+          var presence = {
+            type: "presence",
+            userID: userID,
+            timestamp: data["l"],
```
maerlyn: I haven't seen this field in any of my orca_presence messages, but there's one called "c" that's still a mistery to me.

7osCraft:
@lequanglam  It's fixed in the last commit. Please update

src/listenMqtt.js
7osCraft: Yeah I'm trying to figure out what that is for. I'll update if I found something.

7osCraft:
Fixed ForcedFetch issues and converted presence timestamps from seconds to milliseconds.
You can merge now. :smiley:

Schmavery:
Thanks :)

thanhdatpd:
Hi @7osCraft 
Sometime mqtt client has this error or after call api.logout()
image
listenMqtt just show it and nothing trigger. Can you add some err return like this to handle easy.
image
Thanks and sorry for my English.

7osCraft:
@thanhdatpd Sure. Thanks!

Issue #763, opened 2019-11-17
Listen issues


Alcap:
Hi,

Did anyone notice some change in behaviour on facebook's end? I have an api.listen on my code and it was working just fine and all of the sudden it stopped.

Any code that that I place in the callback doesn't get executed when a message is received. This code was working fine just a few days ago and now, nothing.
There's no errors provided, so I'm at a loss on how to go on the troubleshooting and how to fix this.

Thanks!

Alcap:
Actually, I've tried to pick up the example code and run it as vanilla as possible and it is still unable to react when someone sends a message to any thread.
I've looked up how to increase verbosity of the listen functions, but couldn't find anything of relevance... Any tips?

Thanks

AstroCB:
I'm also experiencing this issue – my listen handler is no longer receiving messages when they are sent.

lequanglam:
Facebook Messenger is using WebSocket now.
![image](https://user-images.githubusercontent.com/29724654/69214326-38426a80-0b99-11ea-858e-2be2b6088762.png)
great, now we need to rewrite that `listen.js` file

7osCraft:
I have the same issue too. I think they now use websockets only.

7osCraft:
This might help too: https://github.com/tulir/fbchat-asyncio/blob/yet-another-rewrite/fbchat/_client.py#L2749-L2882

ccfiel:
> This might help too: https://github.com/tulir/fbchat-asyncio/blob/yet-another-rewrite/fbchat/_client.py#L2749-L2882

I did try this one but still the same problem

seakBz:
I have the same issue too.

KMatuszak:
I also have the same issue.

Schmavery:
If anyone has time, they could experiment with hooking up something like https://www.npmjs.com/package/mqtt maybe

7osCraft:
@Schmavery 
Actually I did. It worked but the library itself needs some modification to work with the facebook protocol.

7osCraft:
I'll try to get a PR ready when I can.

PixelHir:
Stopped working today

PixelHir:
for now i think that [this](https://github.com/ChatPlug/libfb-js)/ could work as a replacement until the issue is resolved

7osCraft:
PR is ready. Waiting to be merged.

KMatuszak:
@7osCraft I have this error on login:

```
[...]

Issue #742, opened 2019-08-06
Facebook unusual login


hambog00:
Facebook provides me a notification on attempt to login:
We noticed an unusual login from a device or location you don't usually use. Was this yo?

my app is running on my local machine..

ravkr:
They are recognizing that you are using facebook-chat-api :/ it's a problem known for at least one year
you might try to change user-agent etc. but I'm not sure if it would fix it

Alcap:
Is there any workaround to this?

lequanglam:
> Is there any workaround to this?

change the user agent, send markAsRead every time receive messages.

Alcap:
Thanks! Will do!

prameshbajra:
> Is there any workaround to this?

Hi @Alcap , were you able find a workaround on this?
Mind guiding me too?

PR #768, opened 2019-12-06
Support for markAsDelivered and auto message delivery and reading.

7osCraft:
Facebook in the browser marks new messages as delivered, So I added that here with the abillity to make it automatic (On by default). Also made an option to make markAsRead automatic but it's not on by default.

I don't know for sure if marking messages as delivered helps with the issue of accounts getting blocked, but it's better to do it anyway.

I fixed an issue I saw in #763 related to listenMqtt too.

Also added the docs to these new changes.
Thanks!

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -1342,7 +1342,7 @@ __Arguments__
 
 Given a threadID and a messageID will mark that message as delivered. If a message is marked as delivered that tells facebook servers that it was recieved.
 
-You can also mark new messages as delivered automatically. See [api.setOptions](#setOptions).
+You can also mark new messages as delivered automatically. This is enabled by default. See [api.setOptions](#setOptions).
 
 __Arguments__
 
--- a/DOCS.md
+++ b/DOCS.md
@@ -27,6 +27,7 @@
 * [`api.listen`](#listen)
 * [`api.listenMqtt`](#listenMqtt)
 * [`api.logout`](#logout)
+* [`api.markAsDelivered`](#markAsDelivered)
 * [`api.markAsRead`](#markAsRead)
 * [`api.markAsReadAll`](#markAsReadAll)
 * [`api.muteThread`](#muteThread)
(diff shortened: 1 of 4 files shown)


Schmavery:
@7osCraft thanks a lot, I will review this soon.

@lequanglam you said [here](https://github.com/Schmavery/facebook-chat-api/issues/742#issuecomment-548788030) that you got blocked less when you mark as read? Should we turn this on by default?

DanH42:
I use a custom user agent (pulled from my own web browser and updated periodically) but I don't automatically mark all messages as read, and I haven't been flagged in months. Instantly marking every single message as read 24/7 also sounds like bot behavior, so it's not something I'd want to happen by default.

I'll also mention that keeping track of appState made a huge difference in preventing security checkpoints, so I'd encourage anyone who's not using it to do so.

7osCraft:
Yes, auto marking as seen might be bot behaviour so that's why I made it false by default, but marking as delivered is done automatically by the browser or the app even if you leave it. Unless you go offline of course.

It also looks like some values in appState changes over time so I think we should do something about it.

7osCraft:
Another thing I do is sending typing indicators before every message. I might add that to the PR.

Schmavery:
@7osCraft if you do that can you make it a separate PR please <3

7osCraft:
Sure 😃

Schmavery:
I agree it would be cool to look into making appState easier to use or at least improving the docs around it so that it's less confusing

DOCS.md
```diff
@@ -1336,11 +1337,46 @@ __Arguments__
 
 ---------------------------------------
 
+<a name="markAsDelivered"></a>
+### api.markAsDelivered(threadID, messageID[, callback]])
+
+Given a threadID and a messageID will mark that message as delivered. If a message is marked as delivered that tells facebook servers that it was recieved.
+
+You can also mark new messages as delivered automatically. See [api.setOptions](#setOptions).
```
Schmavery: Should we mention that this is on by default?

src/listenMqtt.js
```diff
@@ -408,6 +418,67 @@ function parseDelta(ctx, globalCallback, defaultFuncs, v) {
   }
 }
 
+function markDelivery(ctx, defaultFuncs, threadID, messageID) {
```
Schmavery: Why doesn't this use api.markAsDelivered?

DOCS.md
7osCraft: Yes that's a good idea.

src/listenMqtt.js
7osCraft: I could add a require statement with the markAsDelivered and markAsRead.

Schmavery: It should be on the api object, right?

7osCraft: There is no api object inside the api functions

src/listenMqtt.js
```diff
@@ -408,6 +418,67 @@ function parseDelta(ctx, globalCallback, defaultFuncs, v) {
...
+            });
+        }
+      })
+      .catch(function(err) {
+        log.error("markAsDelivered", err);
+      });
+  }
+}
+
 module.exports = function(defaultFuncs, api, ctx) {
```
Schmavery: Here is the api object :)

7osCraft: Oh How did I miss that xD 😂

7osCraft:
Pushed some changes and fixes.

PR #771, opened 2019-12-14
change message_reply senderID type to str

andybui01:
senderID for event type - message is a string. For consistency purposes, message_reply's senderID should be a string and not an integer.

### DIFF
--- a/src/listen.js
+++ b/src/listen.js
@@ -355,7 +355,7 @@ module.exports = function(defaultFuncs, api, ctx) {
                               delta.deltaMessageReply.message.messageMetadata.threadKey.threadFbId : delta.deltaMessageReply.message.messageMetadata.threadKey
                               .otherUserFbId,
                             messageID: delta.deltaMessageReply.message.messageMetadata.messageId,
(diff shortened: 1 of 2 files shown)


andybui01:
Hmm I'm not sure where else senderID is an integer but it's also an integer for "message_reaction" events. In that case, is message's senderID supposed to be an integer?

7osCraft:
It should be a string along with threadID in every event.

Schmavery:
Seems ok to me -- @andybui01 can you fix it for message_reaction too? thanks

andybui01:
I took a closer look at some of the events and here are some events with integer IDs. I'm guessing all of these should be strings instead.

**presence**
- userID

**message_reply**
- threadID
- senderID
- messageReply:
- - threadID
- - senderID

**message_reaction**
- threadID
- senderID
- userID

**message_unsend**
- threadID
- senderID

Schmavery:
Hey @andybui01 sorry about the miscommunication and thanks for bringing up this issue. I think some of the issues may have been addressed in the recently-merged #779 but if there's anything missing feel free to make it known.

PR #778, opened 2020-01-23
Bug fixes. closes #775

### LINKED ISSUE #775: MQTT bug?
image
@7osCraft

7osCraft:
This fixes some errors with mqtt. I also made the mqtt connecting a little better.
I would like to ask If I should do #771 too.
Another thing is that I want to refactor some stuff to support new events better and I want to do something like #628 to make everything clear and better.
Thanks!


Schmavery:
Thanks for doing this, looks good. The other stuff sounds ok if we can do it cleanly enough

PR #779, opened 2020-01-24
Fixes for listenMqtt

lequanglam:
- Change ID type from `number` to `string` (Schmavery#771)
- Fix some perfomance issues (if callback takes a very longgggggggg time to resolve, it will lead to disconnection of MQTT) => wrap callback in a function so it will not block other calls to callback


Schmavery:
Ok thanks

PR #785, opened 2020-01-31
Update for thread colors

### LINKED ISSUE #784: Support for new colors
Can we add support for setting new colors?

BadAimWeeb:
Resolves #784


src/threadColors.js
```diff
@@ -1,22 +1,39 @@
...
+    ////PictonBlue: "#6699cc",
+    ////FreeSpeechGreen: "#13cf13",
+    ////Pumpkin: "#ff7e29",
+    ////LightCoral: "#e68585",
+    ////MediumSlateBlue: "#7646ff",
+    ////DeepSkyBlue: "#20cef5",
+    ////Fern: "#67b868",
+    ////Cameo: "#d4a88c",
+    ////BrilliantRose: "#ff5ca1",
+    ////BilobaFlower: "#a695c7"
```
Schmavery: Are these the same colors with different names? This is a really breaking change...

Everything else looks great

lequanglam: umm, those are some old color which have been commented. also facebook uses theme id now, so those old colors can't be used anymore. 
and no, we can't keep them.

Schmavery: I understand that the api has changed, but clearly some of these colors must be similar `MessengerBlue` vs `DefaultBlue` for example.. I remember the original names came from the messenger code somewhere, where did you get the new names?
I'm sure you understand that this change break anyone who currently uses this api

lequanglam: I get these names from this
image
and yes, I understand that this change will break compatibility.

ravkr: @lequanglam you should put new ids in place of old hex colours (instead of commenting them). This way we can maintain backward compatibility.

lequanglam: @ravkr will try to make it backward-compatible, but i'm currently busy at listenMqtt's bugs.

PR #789, opened 2020-02-15
Made MQTT connect code simpler

ptrcnull:
Removed unused code, eg. detecting protocols or parsing URL just to serialize it later


Hongarc:
You must remove `require fbconnect`

Schmavery:
Thanks for reviewing, @Hongarc 
@7osCraft since you wrote the original version of this (that's being partially deleted here), do you mind taking a quick look and giving your thoughts?

7osCraft:
I took a look and everything looks good. That code was only used to get the protocol but it's not needed anymore.

Schmavery:
Thanks everyone

Issue #801, opened 2020-03-24
Listen can be run only once for an api instance


zanstro:
After I upgraded from 1.6.1 to 1.7.0 I am not able to run api.listenMqtt() again for the same api instance. I use login() to create an api instance and then I export this instance and use it in my code to listen for Facebook messages based on specific requests I receive (I have a http.createServer listening for requests and I use these requests to send messages to FB with api.sendMessage()). 

After each response from my server, I run stopListening() as I don't wait for any more messages from Facebook. Everything works fine for the first request but when I get a second request api.listenMqtt() is not working.

KMatuszak:
I think that these variables should be resetted after stopListening
```js
var identity = function () {};
var mqttClient = undefined;
```

zanstro:
Any idea of how can I access these variables through the api instance? Thank you in advance

KMatuszak:
try to modify listenMqtt in your node_modules

zanstro:
Thank you. I tried it but it didn't help :/

stergios-tegos:
I have the same issue with listenMqtt :( 
Any idea how this could be resolved?
Thanks in advance

returnmk:
This is a major issue for me too. Please advice.

Schmavery:
@7osCraft can we add the mqttClient variable to the API context instead of making it a global variable? sorry I missed this during review

https://github.com/Schmavery/facebook-chat-api/blob/3227d2ef8b8354bcb0f3eab5d293421d4c85f524/index.js#L67-L75

7osCraft:
Sure. I'll look into it and do some tests. Thanks for notifying me.
I'll try to make a fix soon.

returnmk:
@7osCraft thank you for looking into it!

Do we have any updates on this?

7osCraft:
Hello,
Sorry for the delay, I'll try to get a PR today.

Issue #787, opened 2020-02-03
listenMqtt issues (again)


lequanglam:
![image](https://user-images.githubusercontent.com/29724654/73645348-b21ec600-46a9-11ea-83e7-abf5a0ae3c3e.png)
and this thing
![image](https://user-images.githubusercontent.com/29724654/73656571-05514280-46c3-11ea-8c00-250000cb0916.png)
_yes i'm using my own repo, but error part is identical._

Schmavery:
cc @7osCraft

7osCraft:
I think there is a problem with the formatting of ForcedFetch responses, but it should be the same kind of message in ``getThreadHisrory``, I'll look into it.
And for the second image, are you using pages or a normal account?

lequanglam:
normal account.

lequanglam:
Update: Data inside image 1:
![image](https://user-images.githubusercontent.com/29724654/74588191-e345b180-502c-11ea-89d6-16d59a70b744.png)
look like we need to check if it's null.

PR #803, opened 2020-03-31
Fix some issues in MQTT.

7osCraft:
Fixed listenMqtt not working if stopped and ran again from the same instance. (#801).
Also fixed an error with thread images being sometimes undefined. (#787).

### DIFF
--- a/src/listenMqtt.js
+++ b/src/listenMqtt.js
@@ -6,7 +6,6 @@ var mqtt = require('mqtt');
 var websocket = require('websocket-stream');
 
 var identity = function () {};
-var mqttClient = undefined;
 
 var lastSeqId = 0;
 var syncToken;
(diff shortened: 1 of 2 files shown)


src/listenMqtt.js
```diff
@@ -502,6 +502,10 @@ module.exports = function (defaultFuncs, api, ctx) {
   return function (callback) {
     globalCallback = callback;
 
+    //Reset some stuff
+    lastSeqId = 0;
```
Schmavery: Should these also be in ctx?

7osCraft: They are mqtt specific, So I just left them here and reset them when listenMqtt is called and they worked fine.

Schmavery: I can't imagine the mqtt clients are supposed to share this information though

Schmavery: The use case that we're fixing is people who are logging into multiple accounts at the same time.
In this case, there would be one ctx object per "api" object, so if you call listenMqtt on account2 after calling it on account1, it won't reset the values for account1.

If we store the variables globally at the top level of this file, all instances of the api will need to share the values. We don't want any interaction at all between instances of the api.

7osCraft: Alright. Sorry for misunderstanding 😁
I'll change it as soon as I can.

Schmavery:
Looks good, thanks a lot!

PR #811, opened 2020-04-17
Added support for sending and receiving static location attachments

KovalRomanK:
Added a new message type of location and a new attachment of type location as well. You can send static locations (not the live ones) using a latitude and longitude value.


ravkr:
I've not tested this yet, but the code looks fine.

PR #834, opened 2020-08-08
Added null check on sticker pack object

nlwstein:
This change fixed a crash on a thread I was enumerating. Let me know if you need further proof / documentation.


Schmavery:
looks simple/safe, thanks

PR #839, opened 2020-10-14
Remove validation check for reacts

AstroCB:
Facebook changed the backend to allow arbitrary emojis to be passed as reactions to messages. These changes can be seen already via `listen`, but can't be sent from the API because we do a check against what were previously the only valid emojis allowed in the call.

This PR just removes that validation so that any arbitrary reacts can be set on a message via the API.

(Tested it manually and it works)


Schmavery:
thanks

PR #844, opened 2020-11-07
Handle when delta does not contain attachments property

ngocdon0127:
Sometimes, with sticker message in group, delta does not contain 'attachments' property.


Schmavery:
thanks

PR #853, opened 2020-12-05
Add listenMqtt to README.md

ryanmohta:
The method list in the `DOCS.md` file contains a link to the `listenMqtt` method, but the method list in the main `README.md` file does not include this. As a result, it is unclear that this method is available to someone browsing through the documentation from the readme — for instance, as I was using the API, I only saw the readme initially and assumed the `listen` method was the only way to listen for messages.


Schmavery:
sgtm

PR #865, opened 2021-02-05
Add `getMessage` to look up messages by mid

### LINKED ISSUE #850: Any way to query messages by messageID?
I'm looking to retrieve attachments for a message after it's been sent, and I was hoping to do this when a reaction is received on that message, but I'm not sure that's possible.

The `message_reaction` message type only gives you the original message's ID as a reference point (unlike the `message_reply` type, which inlines the entire message object for the replied message).

Is it possible to look up the original message object via its messageID with the existing API? If not, is it feasible to add the entire message object to the `message_reaction` payload the way we do with `message_reply` messages? Or is this something Facebook explicitly includes in the payloads for these message types that we don't have control over?

AstroCB:
This adds a new function `getMessage` to look up a message by its messageID.

Resolves #850

### DIFF
--- a/src/getMessage.js
+++ b/src/getMessage.js
@@ -1,0 +1,79 @@
+"use strict";
+
+var utils = require("../utils");
+var log = require("npmlog");
+
+module.exports = function(defaultFuncs, api, ctx) {
+    return function getMessage(threadID, messageID, callback) {
+      if (!callback) {
+        throw { error: "getMessage: need callback" };
+      }
+
+      if (!threadID || !messageID) {
+        throw { error: "getMessage: need threadID and messageID" };
+      }
+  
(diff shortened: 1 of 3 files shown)


src/getMessage.js
```diff
@@ -0,0 +1,79 @@
...
+                        x.type = "unknown";
+                    }
+                    return x;
+                }),
+                body: fetchData.message.text,
+                mentions: fetchData.message.ranges,
+                timestamp: fetchData.timestamp_precise,
+                messageReply: fetchData.replied_to_message,
+                raw: fetchData,
+            }); })();
```
AstroCB, line 70: Let me know if there's a more structured way to parse the response into a message object.

PR #875, opened 2021-03-30
Change the way of checking if an ID is a user or a group when sending messages

KMatuszak:
Calling many times getUserInfo very quickly can block this function and then sending messages becomes impossible, this way we're not calling this function when sending messages. Tested in groups and private messaging and it does work.


Schmavery (approved):
Haha nice

PR #896, opened 2021-12-11
add catch exception to listenMqtt json parser

beanbocchi:
image
Log the **message** variable and see this error
image
This error occurs when successfully logged in (maybe it's presence message type), and this only occurs in some accounts, not every accounts


beanbocchi:
Dupplicate #897

